PR #1, opened 2019-06-15
Example changes to the core XPlat template, with some safe console filters

pgilmorepf:


source/test/TestApp/PlayFabEventTest.cpp
```diff
@@ -257,7 +257,8 @@ namespace PlayFabUnit
 
     void PlayFabEventTest::AddTests()
     {
-#if (!UNITY_IOS && !UNITY_ANDROID) && (!defined(PLAYFAB_PLATFORM_IOS) && !defined(PLAYFAB_PLATFORM_ANDROID))
+		// TODO: Fix whatever limitation causes this test to fail for these platforms
```
Vicodex: >		 [](start = 0, length = 2)

Should be spaces

templates/PlayFab_Api.cpp.ejs
```diff
@@ -118,3 +118,5 @@ namespace PlayFab
 }
 
 #endif
+
+#pragma warning (enable: 4100) // formal parameters are part of a public interface
```
Vicodex: We usually put this line at the top of files

pgilmorepf: It is.
The line at the top disables it, the line at the bottom re-enables it.

PR #4, opened 2019-06-25
Made internal callback declaration C++ standard compliant

Vicodex:
- This is a necessary change that must be in master to enable a possibility of instanced APIs in the other private repo.
- These changes have been used to generate XPlat SDK which then was successfully tested.


Vicodex:
retest

Vicodex:
retest

PR #7, opened 2019-06-27
Added support of instanced APIs for PS4

Vicodex:
The SDK for PS4 was generated and tested with these changes.


Vicodex:
The second test on Switch failed due to PlayFab server connectivity issues. It passed on the second retry with the same input parameters.

Vicodex:
retest

Vicodex:
retest

PR #8, opened 2019-06-28
[ATG] XPlat: fix compilation warnings (get rid of them where possible)

JTonoJimenez:


source/code/source/playfab/QoS/PlayFabQoSApi.cpp
```diff
@@ -41,7 +41,7 @@ namespace PlayFab
             }
         }
 
-        void OnWriteEventsResult(int httpCode, std::string result, std::shared_ptr<CallRequestContainerBase> reqContainer)
+        void OnWriteEventsResult(int, std::string result, std::unique_ptr<CallRequestContainerBase> reqContainer)
```
Vicodex: We changed it to shared_ptr, so this will need to be a shared_ptr

toddbellMSFT: I didn't see your name on the git blame, so I thought it wasn't that particular change. Thanks for pointing it out

toddbellMSFT:
retest

PR #12, opened 2019-07-18
Bug fix: Property names should be used when reading inputs, just like when writing outputs

SK83RJOSH:
Due to incorrect usage of `getPropertySafeName` some input data is lost. This most notably affects the `GetMatch` function, which has the property `ServerDetails`, those details are lost due to this bug -- thus making Matchmaking with Servers 2.0 impossible when using XPlatCore.


toddbellMSFT:
retest

toddbellMSFT:
The previous tests failed a multi-user login declaring there may have been multiple requests happening at once.

PR #22, opened 2019-10-08
Updating winhttp

toddbellMSFT:
enabling the bility to set a request ID so responses can keep track of what request id this particular callback is for.

### DIFF
--- a/source/code/source/playfab/PlayFabWinHttpPlugin.cpp
+++ b/source/code/source/playfab/PlayFabWinHttpPlugin.cpp
@@ -156,7 +156,13 @@ namespace PlayFab
 
             // Use WinHttpOpen to obtain a session handle
             hSession = WinHttpOpen(L"PlayFab Agent",
+#ifdef PLAYFAB_WIN7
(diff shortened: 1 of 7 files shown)


source/code/source/playfab/PlayFabWinHttpPlugin.cpp
```diff
@@ -277,9 +283,56 @@ namespace PlayFab
...
+                                            &wideRequestIdByteCount,
+                                            WINHTTP_NO_HEADER_INDEX);
+
+                                        std::unique_ptr<wchar_t[]> wideRequestIdBuffer;
+                                        DWORD queryError = GetLastError();
+                                        if (queryError == ERROR_INSUFFICIENT_BUFFER)
+                                        {
+                                            // WinHttpQueryHeaders excludes the null terminator from the length
+                                            DWORD wideRequestIdLength = wideRequestIdByteCount / sizeof(wideRequestIdBuffer[0]);
+                                            wideRequestIdBuffer = std::make_unique<wchar_t[]>(wideRequestIdLength + 1);
```
Vicodex: >std::make_unique [](start = 66, length = 16)

This is available only in C++ 14. I thought we want to support C++ 11? If yes, please consider replacing it with a new equivalent

PR #23, opened 2019-10-11
Toddbell/android fixes

toddbellMSFT:
Fixing some android and iOS issues. There were reports of enum's that weren't enum class's which would throw warnings. There should now be only expected warnings about testTitleData.json files existing or not.

### DIFF
--- a/source/code/include/playfab/PlayFabIOSHttpPlugin.h
+++ b/source/code/include/playfab/PlayFabIOSHttpPlugin.h
@@ -43,7 +43,7 @@ namespace PlayFab
 
             bool Initialize(std::unique_ptr<CallRequestContainerBase>& requestContainer);
             
-            enum State:int
+            enum class State:int
             {
                 None = 0,
                 Pending = RequestTask::None,
--- a/source/code/include/playfab/PlayFabIOSHttpPlugin.h
+++ b/source/code/include/playfab/PlayFabIOSHttpPlugin.h
@@ -45,7 +45,7 @@ namespace PlayFab
             
             enum class State:int
             {
-                None,
+                None = 0,
                 Pending = (int)RequestTask::State::None,
                 Requesting,
                 Finished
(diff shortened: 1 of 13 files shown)


source/code/include/playfab/PlayFabIOSHttpPlugin.h
```diff
@@ -43,10 +43,10 @@ namespace PlayFab
 
             bool Initialize(std::unique_ptr<CallRequestContainerBase>& requestContainer);
             
-            enum State:int
+            enum class State:int
             {
                 None = 0,
-                Pending = RequestTask::None,
+                Pending = (int)RequestTask::State::None,
```
PushpadantK: >Pending [](start = 16, length = 7)

is this an existing bug, none and pending were both 0?

source/test/TestApp/TestReport.cpp
```diff
@@ -81,6 +81,12 @@ namespace PlayFabUnit
             case TestFinishState::PASSED: internalReport.passed += 1; break;
             case TestFinishState::FAILED: internalReport.failures += 1; break;
             case TestFinishState::SKIPPED: internalReport.skipped += 1; break;
+            case TestFinishState::PENDING:
```
MichaelMackend: Why are we handling these two states with 'break' instead of letting default handle them.  If it's for the sake of handling all possible states, default should probably log or assert as it would be handling a completely invalid value.

toddbellMSFT: the warning on iOS was that there were defined enums that weren't being caught. I was considering letting each fall through to default, but recalled one compiler threw a warning when it came across the fall through. This way we don't see any warnings, and the compiler should be optimizing these empty statements away.

source/code/include/playfab/PlayFabIOSHttpPlugin.h
toddbellMSFT: It's not currently logged as a bug, should we consider it one?

PR #24, opened 2019-10-22
Toddbell/const fixes

toddbellMSFT:
W.R.T. https://dev.azure.com/playfab/PlayFab/_sprints/taskboard/Tools%20and%20SDK/PlayFab/2019/1911/Sprint%2022%20(Ends%2011-1)?workitem=33113 

I have been combing through our C++ code base attempting to update const correctness where possible. There's a few places where we weren't changing an object at all, but we were moving entire references to that object around just to read a single const variable.

### DIFF
--- a/source/code/include/playfab/PlayFabAndroidHttpPlugin.h
+++ b/source/code/include/playfab/PlayFabAndroidHttpPlugin.h
@@ -36,7 +36,7 @@ namespace PlayFab
         virtual void SetPredefinedHeaders(const RequestTask& requestTask);
         virtual void SetHeader(const RequestTask& requestTask, const char* name, const char* value);
         virtual bool GetBinaryPayload(RequestTask& requestTask, void*& payload, size_t& payloadSize) const;
-        virtual void ProcessResponse(const RequestTask& requestTask, const int httpCode);
+        virtual void ProcessResponse(RequestTask& requestTask, const int httpCode);
         virtual void HandleResults(RequestTask& requestTask);
 
         struct RequestImpl;
--- a/source/code/include/playfab/PlayFabAndroidHttpPlugin.h
+++ b/source/code/include/playfab/PlayFabAndroidHttpPlugin.h
@@ -32,7 +32,7 @@ namespace PlayFab
         bool CheckResponse(RequestTask& requestTask);
         void SetResponseAsBadRequest(RequestTask& requestTask);
 
-        virtual std::string GetUrl(RequestTask& requestTask) const;
+        virtual std::string GetUrl(const RequestTask& requestTask) const;
         virtual void SetPredefinedHeaders(RequestTask& requestTask);
         virtual void SetHeader(RequestTask& requestTask, const char* name, const char* value);
         virtual bool GetBinaryPayload(RequestTask& requestTask, void*& payload, size_t& payloadSize) const;
--- a/source/code/include/playfab/PlayFabAndroidHttpPlugin.h
+++ b/source/code/include/playfab/PlayFabAndroidHttpPlugin.h
@@ -33,10 +33,10 @@ namespace PlayFab
         void SetResponseAsBadRequest(RequestTask& requestTask);
 
         virtual std::string GetUrl(const RequestTask& requestTask) const;
-        virtual void SetPredefinedHeaders(RequestTask& requestTask);
-        virtual void SetHeader(RequestTask& requestTask, const char* name, const char* value);
+        virtual void SetPredefinedHeaders(const RequestTask& requestTask);
+        virtual void SetHeader(const RequestTask& requestTask, const char* name, const char* value);
         virtual bool GetBinaryPayload(RequestTask& requestTask, void*& payload, size_t& payloadSize) const;
-        virtual void ProcessResponse(RequestTask& requestTask, const int httpCode);
+        virtual void ProcessResponse(const RequestTask& requestTask, const int httpCode);
         virtual void HandleResults(RequestTask& requestTask);
 
         struct RequestImpl;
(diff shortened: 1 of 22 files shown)


source/code/include/playfab/QoS/PlayFabQoSApi.h
```diff
@@ -28,15 +28,15 @@ namespace PlayFab
...
             int SetupSockets(std::vector<std::shared_ptr<QoSSocket>>& sockets, unsigned int numThreads, unsigned int timeoutMs);
             void InitializeAsyncPingResults(std::vector<std::future<PingResult>>& asyncPingResults);
-            void PingServers(std::vector<std::string>& pings, std::vector<std::future<PingResult>>& asyncPingResults, std::vector<std::shared_ptr<QoSSocket>>& sockets, std::unordered_map<std::string, PingResult>& accumulatedPingResults, unsigned int timeoutMs);
-            void UpdateAccumulatedPingResult(PingResult& result, std::string region, std::unordered_map<std::string, PingResult>& accumulatedPingResults, unsigned int timeoutMs);
+            void PingServers(const std::vector<std::string>& pings, std::vector<std::future<PingResult>>& asyncPingResults, const std::vector<std::shared_ptr<QoSSocket>>& sockets, std::unordered_map<std::string, PingResult>& accumulatedPingResults, unsigned int timeoutMs);
+            void UpdateAccumulatedPingResult(const PingResult& result, const std::string& region, std::unordered_map<std::string, PingResult>& accumulatedPingResults, unsigned int timeoutMs);
```
pgilmorepf: Holey carp! Another one! Once again, good catch.

pgilmorepf: Given that we seem to have multiple of these, can you do a regex search for any std::strings that seem to be copied rather than ref'ed?

source/code/include/playfab/QoS/RegionResult.h
```diff
@@ -13,7 +13,7 @@ namespace PlayFab
         struct RegionResult
         {
         public:
-            RegionResult(std::string region, int latencyMs, int errorCode);
+            RegionResult(const std::string& region, int latencyMs, int errorCode);
```
pgilmorepf: Doh man, another one. Yeah, clearly we overlooked a lot.

source/code/source/playfab/QoS/PlayFabQoSApi.cpp
```diff
@@ -59,8 +59,8 @@ namespace PlayFab
 
         void WriteTelemetryEvents(
             WriteEventsRequest& request,
-            ProcessApiCallback<WriteEventsResponse> callback,
-            ErrorCallback errorCallback = nullptr,
+            const ProcessApiCallback<WriteEventsResponse>& callback,
+            const ErrorCallback& errorCallback = nullptr,
```
pgilmorepf: I'm not sure I agree with these...
A pointer is just an int64? There's no real reason to accept these by ref... Except these are std::func...
Ok, please explain the full context here. I think I may be vaguely understanding the point here, but please explain it for me, so I can fully grok.

pgilmorepf: Mostly, I'm concerned about what the caller might do with their version of the std::func, and whether it's safe to keep the obj ref, or safer to copy.

pgilmorepf:
Overall, heartily approved. I feel you caught enough std::string copies that you probably did a good regex search for that, but if you feel there's a more thorough search for string copies, please do that.
Otherwise, there's just the std::func copy vs ref, which I want some extra thought about whether it's safe to keep the ref to the one that the caller created.
Overall, this looks awesome, thanks!

source/code/include/playfab/PlayFabAndroidHttpPlugin.h
```diff
@@ -58,6 +58,10 @@ namespace PlayFab
             {
                 return *dynamic_cast<CallRequestContainer*>(requestContainer.get());
             }
+            std::string GetRequestContainerUrl() const
```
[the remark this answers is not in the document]
playfabjenkinsbotDT: The above is used in ProcessResponse where the returned object gets modified (usually the errorWrapper). The dynamic cast to a pointer just to return what that object points to is an odd roundabout way to get the underlying request container out of the std::shared_ptr.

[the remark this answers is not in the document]
toddbellMSFT: The above is used in ProcessResponse where the returned object gets modified (usually the errorWrapper). The dynamic cast to a pointer just to return what that object points to is an odd roundabout way to get the underlying request container out of the std::shared_ptr.

source/code/source/playfab/QoS/PlayFabQoSApi.cpp
toddbellMSFT: So this change is one I was also a bit on the fence with.

a callback should just be a function pointer, which const & wouldn't do much for

std::function on the other hand has some header info that would come with a more object like param.

From our perspective we will never change that function. BUT would that mean they couldn't do something like pass a lambda with this? (From what I'm seeing in our tests, this solution does still work).

source/code/include/playfab/QoS/RegionResult.h
Vicodex: It was an enum before, then hastily replaced to a string at some point. No wonder we missed it.. 

---
In reply to: [337638194](https://github.com/PlayFab/XPlatCoreTemplate/pull/24#discussion_r337638194) [](ancestors = 337638194)

source/code/source/playfab/QoS/PlayFabQoSApi.cpp
Vicodex: That's a valid concern.. References can be extremely dangerous when we don't control the lifespan of the argument

---
In reply to: [337640191](https://github.com/PlayFab/XPlatCoreTemplate/pull/24#discussion_r337640191) [](ancestors = 337640191)

PR #28, opened 2019-10-25
Removes most of the warnings in XPlat core and XPlat consoles

pgilmorepf:
Out of order initialization in some constructors
FastWriter is deprecated, and easily replaced with toStyledString
4100 warnings only apply to Windows
Removing other pragma disables that aren't relevant anymore
And one bit of missed const correctness

### DIFF
--- a/source/code/source/playfab/PlayFabWinHttpPlugin.cpp
+++ b/source/code/source/playfab/PlayFabWinHttpPlugin.cpp
@@ -6,7 +6,6 @@
 #include <vector>
 #include <windows.h>
 
-#pragma warning (disable: 4100) // formal parameters are part of a public interface
 #pragma warning (disable: 4245) // Some DWORD arguments of WinHTTP API can accept -1, according to documentation
 
 namespace PlayFab
--- a/source/code/source/playfab/PlayFabWinHttpPlugin.cpp
+++ b/source/code/source/playfab/PlayFabWinHttpPlugin.cpp
@@ -1,5 +1,7 @@
 #include <stdafx.h>
 
+#if defined(PLAYFAB_PLATFORM_WINDOWS)
+
 #include <playfab/PlayFabWinHttpPlugin.h>
 #include <playfab/PlayFabSettings.h>
 #include <stdexcept>
@@ -191,7 +193,7 @@ namespace PlayFab
                         auto headers = reqContainer.GetHeaders();
                         if (headers.size() > 0)
                         {
-                            for (auto const &obj : headers)
+                            for (auto const& obj : headers)
                             {
                                 if (obj.first.length() != 0 && obj.second.length() != 0) // no empty keys or values in headers
                                 {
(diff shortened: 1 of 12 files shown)


source/code/source/playfab/QoS/PlayFabQoSApi.cpp
```diff
@@ -66,9 +66,7 @@ namespace PlayFab
         {
             IPlayFabHttpPlugin& http = *PlayFabPluginManager::GetPlugin<IPlayFabHttpPlugin>(PlayFabPluginContract::PlayFab_Transport);
             const auto requestJson = request.ToJson();
-
-            Json::FastWriter writer;
```
toddbellMSFT: YES! This will help clear out a lot of issues (may also get rid of the iOS bug https://dev.azure.com/playfab/PlayFab/_workitems/edit/37125 )

templates/PlayFab_Api.cpp.ejs
```diff
@@ -29,17 +31,15 @@ namespace PlayFab
 <% for (var callIdx = 0; callIdx < api.calls.length; callIdx++) { var apiCall = api.calls[callIdx]; %>
     void PlayFab<%- api.name %>API::<%- apiCall.name %>(
         <%- apiCall.request %>& request,
-        ProcessApiCallback<<%- apiCall.result %>> callback,
-        ErrorCallback errorCallback,
+        const ProcessApiCallback<<%- apiCall.result %>>& callback,
```
toddbellMSFT: Oo, make sure these are non const&, you may want to double check that you merge with master

pgilmorepf: Ok, I think const is right, but Deb and Victor convinced me that & is bad.

source/code/source/playfab/PlayFabWinHttpPlugin.cpp
```diff
@@ -365,6 +367,9 @@ namespace PlayFab
 
     bool PlayFabWinHttpPlugin::GetBinaryPayload(CallRequestContainer& reqContainer, LPVOID& payload, DWORD& payloadSize) const
     {
+        UNREFERENCED_PARAMETER(reqContainer);
```
toddbellMSFT: Perfect, I remember seeing some errors on the mac regarding unreferenced params.

source/code/source/playfab/QoS/PlayFabQoSApi.cpp
```diff
@@ -59,16 +59,14 @@ namespace PlayFab
 
         void WriteTelemetryEvents(
             WriteEventsRequest& request,
-            const ProcessApiCallback<WriteEventsResponse>& callback,
-            const ErrorCallback& errorCallback = nullptr,
+            const ProcessApiCallback<WriteEventsResponse> callback,
```
toddbellMSFT: These still don't need to be const.

templates/PlayFab_API.h.ejs
```diff
@@ -22,7 +22,7 @@ namespace PlayFab
 <% } %>
         // ------------ Generated API calls
 <% for (var callIdx = 0; callIdx < api.calls.length; callIdx++) { var apiCall = api.calls[callIdx];
-%>        static void <%- apiCall.name %>(<%- api.name %>Models::<%- apiCall.request %>& request, ProcessApiCallback<<%- api.name %>Models::<%- apiCall.result %>> callback, ErrorCallback errorCallback = nullptr, void* customData = nullptr);
+%>        static void <%- apiCall.name %>(<%- api.name %>Models::<%- apiCall.request %>& request, const ProcessApiCallback<<%- api.name %>Models::<%- apiCall.result %>> callback, const ErrorCallback errorCallback = nullptr, void* customData = nullptr);
```
toddbellMSFT: I didn't think these pass by value functions should be const

source/code/source/playfab/QoS/PlayFabQoSApi.cpp
pgilmorepf: Any reason they shouldn't be?

templates/PlayFab_API.h.ejs
pgilmorepf: Any reason they shouldn't be?
We won't modify their function pointers.

PR #29, opened 2019-10-25

toddbellMSFT:
This was causing a SetMemory issue


pgilmorepf:
So, I'm thinking it probably doesn't actually affect anything, but makes it syntactically nicer, or better in the debugger or something?
What's the goal here?

PR #33, opened 2019-10-31
Toddbell/auto fix

toddbellMSFT:
Updating Auto correctness standards in our codebase.

Auto may be used in only 1 of 3 cases
    1.) the Class is directly being called out/allocated/what not on the right hand side of the equals sign
        NOTE: This does NOT include Getter functions/Builder functions. There are cases where getter functions may reference a more general object in the method name, rather than the specific (i.e. PlayFabSettings or even PlayFabSharedSettings instead of GetSettings).
    2.) iterators (may be const & or not, depends on what you are doing with this iterator)
    3.) binds (prefer const auto & for these unless you need to be changing a bind on the fly, but that sounds like a discussion for another time)

EDIT: 
    This also contains some type changes that popped up extra warnings on other platforms (mostly either Android Studio or Xcode marking deprecation's as warnings).

### DIFF
--- a/source/code/source/playfab/PlayFabAndroidHttpPlugin.cpp
+++ b/source/code/source/playfab/PlayFabAndroidHttpPlugin.cpp
@@ -431,7 +431,7 @@ namespace PlayFab
 
         // Call SetUrl
         {
-            const std::string& requestUrl = GetUrl(requestTask);
+            const std::string requestUrl = GetUrl(requestTask);
 
             jmethodID methodId = jniEnv->GetMethodID(GetHelper().GetHttpRequestClass(), "setUrl", "(Ljava/lang/String;)Z");
             if (methodId == nullptr)
@@ -456,7 +456,8 @@ namespace PlayFab
         SetPredefinedHeaders(requestTask);
 
         // Call SetHeader
-        const auto& headers = requestContainer.GetRequestHeaders();
+        const std::unordered_map<std::string, std::string> headers = requestContainer.GetRequestHeaders();
+        
         if (!headers.empty())
         {
             for (auto const &obj : headers)
--- a/source/code/source/playfab/PlayFabAndroidHttpPlugin.cpp
+++ b/source/code/source/playfab/PlayFabAndroidHttpPlugin.cpp
@@ -431,7 +431,7 @@ namespace PlayFab
 
         // Call SetUrl
         {
-            auto requestUrl = GetUrl(requestTask);
+            const std::string& requestUrl = GetUrl(requestTask);
 
             jmethodID methodId = jniEnv->GetMethodID(GetHelper().GetHttpRequestClass(), "setUrl", "(Ljava/lang/String;)Z");
             if (methodId == nullptr)
@@ -456,7 +456,7 @@ namespace PlayFab
         SetPredefinedHeaders(requestTask);
 
         // Call SetHeader
-        auto headers = requestContainer.GetRequestHeaders();
+        const auto& headers = requestContainer.GetRequestHeaders();
         if (!headers.empty())
         {
             for (auto const &obj : headers)
@@ -470,7 +470,7 @@ namespace PlayFab
 
         // Call SetBody
         {
-            auto requestUrl = GetUrl(requestTask);
+            const std::string& requestUrl = GetUrl(requestTask);
 
             jmethodID methodId = jniEnv->GetMethodID(GetHelper().GetHttpRequestClass(), "setBody", "([B)V");
             if (methodId == nullptr)
@@ -689,7 +689,7 @@ namespace PlayFab
     {
         CallRequestContainer& requestContainer = requestTask.RequestContainer();
 
-        auto callback = requestContainer.GetCallback();
+        const auto& callback = requestContainer.GetCallback();
         if (callback != nullptr)
         {
             callback(requestContainer.responseJson.get("code", Json::Value::null).asInt(),
--- a/source/code/source/playfab/PlayFabAndroidHttpPlugin.cpp
+++ b/source/code/source/playfab/PlayFabAndroidHttpPlugin.cpp
@@ -561,7 +561,7 @@ namespace PlayFab
                 methodId = jniEnv->GetMethodID(GetHelper().GetHttpRequestClass(), "getResponseHttpBody", "()[B");
                 if (methodId)
                 {
-                    auto responseBody = (jbyteArray)jniEnv->CallObjectMethod(httpRequestObject, methodId);
+                    jbyteArray responseBody = (jbyteArray)jniEnv->CallObjectMethod(httpRequestObject, methodId);
 
                     if (responseBody != nullptr)
                     {
--- a/source/code/source/playfab/PlayFabAndroidHttpPlugin.cpp
+++ b/source/code/source/playfab/PlayFabAndroidHttpPlugin.cpp
@@ -471,7 +471,7 @@ namespace PlayFab
 
         // Call SetBody
         {
-            const std::string& requestUrl = GetUrl(requestTask);
+            const std::string requestUrl = GetUrl(requestTask);
 
             jmethodID methodId = jniEnv->GetMethodID(GetHelper().GetHttpRequestClass(), "setBody", "([B)V");
             if (methodId == nullptr)
@@ -690,7 +690,7 @@ namespace PlayFab
     {
         CallRequestContainer& requestContainer = requestTask.RequestContainer();
 
-        const auto& callback = requestContainer.GetCallback();
+        CallRequestContainerCallback callback = requestContainer.GetCallback();
         if (callback != nullptr)
         {
             callback(requestContainer.responseJson.get("code", Json::Value::null).asInt(),
--- a/source/code/source/playfab/PlayFabCallRequestContainer.cpp
+++ b/source/code/source/playfab/PlayFabCallRequestContainer.cpp
@@ -24,14 +24,12 @@ namespace PlayFab
         Json::Value request;
         std::string errs;
         Json::CharReaderBuilder builder;
-        Json::CharReader* reader = builder.newCharReader();
+        std::unique_ptr<Json::CharReader> reader(builder.newCharReader());
 
         const char* reqBod = requestBody.c_str();
         size_t reqBodLength = requestBody.length();
 
         bool parsingSuccessful = reader->parse(reqBod, reqBod + reqBodLength, &request, &errs);
-
-        delete reader;
 
         if (parsingSuccessful)
         {
--- a/source/code/source/playfab/PlayFabCallRequestContainer.cpp
+++ b/source/code/source/playfab/PlayFabCallRequestContainer.cpp
@@ -5,9 +5,9 @@
 
 namespace PlayFab
 {
-    CallRequestContainer::CallRequestContainer(std::string url,
+    CallRequestContainer::CallRequestContainer(const std::string& url,
         const std::unordered_map<std::string, std::string>& headers,
-        std::string requestBody,
+        const std::string& requestBody,
         CallRequestContainerCallback callback,
         void* customData,
         std::shared_ptr<PlayFabApiSettings> settings) :
@@ -22,8 +22,15 @@ namespace PlayFab
         errorWrapper.UrlPath = url;
 
         Json::Value request;
-        Json::Reader reader;
-        bool parsingSuccessful = reader.parse(requestBody, request);
+        std::string errs;
+        Json::CharReaderBuilder builder;
+        Json::CharReader* reader = builder.newCharReader();
+
+        const char* reqBod = requestBody.c_str();
+        size_t reqBodLength = requestBody.length();
+
+        bool parsingSuccessful = reader->parse(reqBod, reqBod + reqBodLength, &request, &errs);
+
 
         if (parsingSuccessful)
         {
--- a/source/code/source/playfab/PlayFabCallRequestContainer.cpp
+++ b/source/code/source/playfab/PlayFabCallRequestContainer.cpp
@@ -31,6 +31,7 @@ namespace PlayFab
 
         bool parsingSuccessful = reader->parse(reqBod, reqBod + reqBodLength, &request, &errs);
 
+        delete reader;
 
         if (parsingSuccessful)
         {
--- a/source/code/source/playfab/PlayFabCurlHttpPlugin.cpp
+++ b/source/code/source/playfab/PlayFabCurlHttpPlugin.cpp
@@ -128,7 +128,7 @@ namespace PlayFab
         curlHttpHeaders = curl_slist_append(curlHttpHeaders, ("X-PlayFabSDK: " + PlayFabSettings::versionString).c_str());
         curlHttpHeaders = curl_slist_append(curlHttpHeaders, "X-ReportErrorAsSuccess: true");
 
-        auto headers = reqContainer.GetRequestHeaders();
+        const auto& headers = reqContainer.GetRequestHeaders();
 
         if (headers.size() > 0)
         {
@@ -157,7 +157,7 @@ namespace PlayFab
 
         // Send
         curl_easy_setopt(curlHandle, CURLOPT_SSL_VERIFYPEER, true);
-        const auto res = curl_easy_perform(curlHandle);
+        const auto& res = curl_easy_perform(curlHandle);
         long curlHttpResponseCode = 0;
         curl_easy_getinfo(curlHandle, CURLINFO_RESPONSE_CODE, &curlHttpResponseCode);
 
@@ -209,7 +209,7 @@ namespace PlayFab
     void PlayFabCurlHttpPlugin::HandleResults(std::unique_ptr<CallRequestContainer> requestContainer)
     {
         CallRequestContainer& reqContainer = *requestContainer;
-        auto callback = reqContainer.GetCallback();
+        const auto& callback = reqContainer.GetCallback();
         if (callback != nullptr)
         {
             callback(
--- a/source/code/source/playfab/PlayFabCurlHttpPlugin.cpp
+++ b/source/code/source/playfab/PlayFabCurlHttpPlugin.cpp
@@ -209,7 +209,7 @@ namespace PlayFab
     void PlayFabCurlHttpPlugin::HandleResults(std::unique_ptr<CallRequestContainer> requestContainer)
     {
         CallRequestContainer& reqContainer = *requestContainer;
-        const auto& callback = reqContainer.GetCallback();
+        CallRequestContainerCallback callback = reqContainer.GetCallback();
         if (callback != nullptr)
         {
             callback(
--- a/source/code/source/playfab/PlayFabCurlHttpPlugin.cpp
+++ b/source/code/source/playfab/PlayFabCurlHttpPlugin.cpp
@@ -128,7 +128,7 @@ namespace PlayFab
         curlHttpHeaders = curl_slist_append(curlHttpHeaders, ("X-PlayFabSDK: " + PlayFabSettings::versionString).c_str());
         curlHttpHeaders = curl_slist_append(curlHttpHeaders, "X-ReportErrorAsSuccess: true");
 
-        const auto& headers = reqContainer.GetRequestHeaders();
+        const std::unordered_map<std::string, std::string> headers = reqContainer.GetRequestHeaders();
 
         if (headers.size() > 0)
         {
@@ -157,7 +157,7 @@ namespace PlayFab
 
         // Send
         curl_easy_setopt(curlHandle, CURLOPT_SSL_VERIFYPEER, true);
-        const auto& res = curl_easy_perform(curlHandle);
+        CURLcode res = curl_easy_perform(curlHandle);
         long curlHttpResponseCode = 0;
         curl_easy_getinfo(curlHandle, CURLINFO_RESPONSE_CODE, &curlHttpResponseCode);
 
--- a/source/code/source/playfab/PlayFabEventBuffer.cpp
+++ b/source/code/source/playfab/PlayFabEventBuffer.cpp
@@ -35,7 +35,7 @@ namespace PlayFab
 
         // First event is just a stub which is used only to maintain the initial consistency of m_head/m_tail.
         // It is immediately considered already consumed.
-        unsigned long long index = eventIndex->load(std::memory_order_relaxed);
+        auto index = static_cast<uint64_t>(eventIndex->load(std::memory_order_relaxed));
         PlayFabEventPacket* firstEvent = CreateEventPacket(buffer, index, nullptr);
 
         tail = firstEvent;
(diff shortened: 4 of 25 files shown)


source/code/source/playfab/PlayFabAndroidHttpPlugin.cpp
```diff
@@ -431,7 +431,7 @@ namespace PlayFab
 
         // Call SetUrl
         {
-            auto requestUrl = GetUrl(requestTask);
+            const std::string& requestUrl = GetUrl(requestTask);
```
ScottMunroMS: >const std::string& [](start = 12, length = 18)

FWIW calling GetUrl constructs a string so using `auto` here (and similarly for GetRequestHeaders below) doesn't actually result in a string copy and using const std::string& doesn't save us any string constructions. I don't have a preference one way or the other what you do here but it may be misleading to some that the variable is a reference type when GetUrl isn't returning a reference.

ScottMunroMS: also applies to your GetCallback() below since that also returns a Callback value and not a reference 

---
In reply to: [340919974](https://github.com/PlayFab/XPlatCoreTemplate/pull/33#discussion_r340919974) [](ancestors = 340919974)

source/code/source/playfab/PlayFabCallRequestContainer.cpp
```diff
@@ -22,8 +22,16 @@ namespace PlayFab
...
+        std::string errs;
+        Json::CharReaderBuilder builder;
+        Json::CharReader* reader = builder.newCharReader();
+
+        const char* reqBod = requestBody.c_str();
+        size_t reqBodLength = requestBody.length();
+
+        bool parsingSuccessful = reader->parse(reqBod, reqBod + reqBodLength, &request, &errs);
+
+        delete reader;
```
ScottMunroMS: what motivated this change? It doesn't seem to have any relation to avoiding copies or auto

source/code/source/playfab/PlayFabCallRequestContainer.cpp
```diff
@@ -22,8 +22,16 @@ namespace PlayFab
...
```
ScottMunroMS: instead of explicit delete I suggest you manage the reader pointer with a smart pointer (i.e. unique_ptr)

source/code/source/playfab/PlayFabCurlHttpPlugin.cpp
```diff
@@ -157,7 +157,7 @@ namespace PlayFab
 
         // Send
         curl_easy_setopt(curlHandle, CURLOPT_SSL_VERIFYPEER, true);
-        const auto res = curl_easy_perform(curlHandle);
+        const auto& res = curl_easy_perform(curlHandle);
```
ScottMunroMS: >const auto& [](start = 8, length = 11)

curl_easy is a c-api I'm pretty sure which means that there's no circumstance where the value being returned is ever a reference type, so similar to my other comment this doesn't have much value and may be confusing to others.

I'm not here to impose bumblelion style on the playfab codebase, but something we do (which I've really taken a shine to) is to only allow use of auto when the type is clearly present on the right-hand side of the initialization expression

e.g.
`auto x = static_cast<int*>(someVoidPtr);`

but all other places have to use full types

`<the actual curl type is> res = curl_easy_perform(...)`

source/code/source/playfab/PlayFabEventBuffer.cpp
```diff
@@ -35,7 +35,7 @@ namespace PlayFab
 
         // First event is just a stub which is used only to maintain the initial consistency of m_head/m_tail.
         // It is immediately considered already consumed.
-        auto index = eventIndex->load(std::memory_order_relaxed);
+        unsigned long long index = eventIndex->load(std::memory_order_relaxed);
```
ScottMunroMS: >unsigned long long  [](start = 8, length = 19)

eventIndex is a std::shared_ptr<std::atomic<uint64_t>>

comments aside on how necessary that is, that tells me that the type returned by `load` is u64 not `unsigned long long` (they may evaluate to equivalent types on most compilers but it's unnecessarily confusing to change the type here)

source/code/source/playfab/PlayFabEventPipeline.cpp
```diff
@@ -106,7 +106,7 @@ namespace PlayFab
             }
 
             // pipeline failed to intake the event, create a response
-            const auto& playFabEmitRequest = std::dynamic_pointer_cast<const PlayFabEmitEventRequest>(request);
+            std::shared_ptr<const PlayFabEmitEventRequest> playFabEmitRequest = std::dynamic_pointer_cast<const PlayFabEmitEventRequest>(request);
             auto playFabEmitEventResponse = std::shared_ptr<PlayFabEmitEventResponse>(new PlayFabEmitEventResponse());
```
ScottMunroMS: >td::shared_ptr<PlayFabEmitEventResponse>(new [](start = 45, length = 44)

nit: total tangent so feel free to "won't fix" but you could simplify this line as `std::make_shared<PlayFabEmitEventResponse>()`

source/code/source/playfab/PlayFabEventPipeline.cpp
```diff
@@ -272,10 +272,10 @@ namespace PlayFab
                 // call individual emit event callbacks
                 for (const auto& eventEmitRequest : *requestBatchPtr)
                 {
-                    const auto& playFabEmitRequest = std::dynamic_pointer_cast<const PlayFabEmitEventRequest>(eventEmitRequest);
+                    std::shared_ptr<const PlayFabEmitEventRequest> playFabEmitRequest = std::dynamic_pointer_cast<const PlayFabEmitEventRequest>(eventEmitRequest);
                     auto playFabEmitEventResponse = std::shared_ptr<PlayFabEmitEventResponse>(new PlayFabEmitEventResponse());
                     playFabEmitEventResponse->emitEventResult = EmitEventResult::Success;
-                    auto playFabError = std::shared_ptr<PlayFabError>(new PlayFabError());
+                    std::shared_ptr<PlayFabError> playFabError = std::shared_ptr<PlayFabError>(new PlayFabError());
```
ScottMunroMS: >std::shared_ptr<PlayFabError> [](start = 20, length = 29)

you've expanded this type but the `playFabEmitEventResponse` variable 2 lines above which uses the same initialization pattern didn't change. seems inconsistent.

frankly this is an instance where I'd just use auto + make_shared (as alluded to in another comment)

`auto playFabError = std::make_shared<PlayFabError>();`

or if you really want to expand the type you can do...

`std::shared_ptr<PlayFabError> playFabError (new PlayFabError());`

this latter one eliminates the need to put std::shared_ptr on both sides which makes the line easier to read

source/code/source/playfab/PlayFabPluginManager.cpp
```diff
@@ -27,8 +27,8 @@ namespace PlayFab
 
     std::shared_ptr<IPlayFabPlugin> PlayFabPluginManager::GetPluginInternal(const PlayFabPluginContract contract, const std::string& instanceName)
     {
-        const auto key = std::make_pair(contract, instanceName);
-        auto pluginEntry = plugins.find(key);
+        std::pair<PlayFabPluginContract, std::string> key = std::make_pair(contract, instanceName);
+        const auto& pluginEntry = plugins.find(key);
```
ScottMunroMS: >const auto& [](start = 8, length = 11)

similar comment as elsewhere. find doesn't return an reference type. it returns a value type so this may be misleading. My guess is you used auto because the full iterator type was long and ugly. I empathize but don't think using `const auto&` here is better than `auto` or using the full type

source/code/source/playfab/PlayFabWinHttpPlugin.cpp
```diff
@@ -255,7 +255,7 @@ namespace PlayFab
                                         }
 
                                         // Allocate space for the buffer
-                                        auto outBuffer = std::unique_ptr<char[]>(new char[dwSize + 1]);
+                                        std::unique_ptr<char[]> outBuffer = std::unique_ptr<char[]>(new char[dwSize + 1]);
```
ScottMunroMS: >std::unique_ptr<char[]> [](start = 40, length = 23)

the type is clearly stated on the right side. I don't think removing the auto here really helps much. again I'll leave it to your style discretion but I think leaving this as auto would be fine

ScottMunroMS:
oh geez. Apologies. I just realized I totally glossed over your description where you lay out what "auto" rules you wanted to establish. There are a number of comments I've left where I didn't really make a good statement because I wasn't clear on what auto rules you wanted in the codebase. I'll do a pass over all the comments I've previously left and try to align them to the style rules laid out in your description

source/code/source/playfab/PlayFabCurlHttpPlugin.cpp
ScottMunroMS: from your auto rules in the description it sounds like you'd prefer using the full type here is as well since it's not obvious from the right side.

---
In reply to: [340921208](https://github.com/PlayFab/XPlatCoreTemplate/pull/33#discussion_r340921208) [](ancestors = 340921208)

source/code/source/playfab/PlayFabEventPipeline.cpp
ScottMunroMS: from your description it sounds like either of these options would more align with the auto rules you are trying to establish. Either using auto because the type is obviously on the right hand side or avoiding the left vs right hand type question entirely

---
In reply to: [340922563](https://github.com/PlayFab/XPlatCoreTemplate/pull/33#discussion_r340922563) [](ancestors = 340922563)

source/code/source/playfab/PlayFabPluginManager.cpp
ScottMunroMS: the rules in your description say "auto" is okay for iterators. I don't mind that but similar to other comments it's worth reconsidering the const ref portion of the type since it doesn't represent the return value of the function being called.

---
In reply to: [340922982](https://github.com/PlayFab/XPlatCoreTemplate/pull/33#discussion_r340922982) [](ancestors = 340922982)

ScottMunroMS:
turns out I hadn't left nearly as many goofy comments as I thought! anyway I finished my quick pass over the old commnets

---
In reply to: [548174317](https://github.com/PlayFab/XPlatCoreTemplate/pull/33#issuecomment-548174317) [](ancestors = 548174317)

source/code/source/playfab/PlayFabAndroidHttpPlugin.cpp
```diff
@@ -456,7 +456,7 @@ namespace PlayFab
         SetPredefinedHeaders(requestTask);
 
         // Call SetHeader
-        auto headers = requestContainer.GetRequestHeaders();
+        const auto& headers = requestContainer.GetRequestHeaders();
```
ScottMunroMS: >auto [](start = 14, length = 4)

it's not obvious what this type is from the right-hand side which seems to contradict your auto rules. I agree with those rules and think the full type is better here (similarly for GetCallback)

source/code/source/playfab/QoS/PlayFabQoSApi.cpp
```diff
@@ -48,10 +48,10 @@ namespace PlayFab
             WriteEventsResponse outResult;
             if (ValidateResult(outResult, container))
             {
-                const auto internalPtr = container.successCallback.get();
+                const auto& internalPtr = container.successCallback.get();
```
ScottMunroMS: >auto [](start = 22, length = 4)

it's not clear from the right-hand side what this type is.
recommend using the full type

source/code/source/playfab/QoS/PlayFabQoSApi.cpp
```diff
@@ -65,13 +65,13 @@ namespace PlayFab
...
             IPlayFabHttpPlugin& http = *PlayFabPluginManager::GetPlugin<IPlayFabHttpPlugin>(PlayFabPluginContract::PlayFab_Transport);
-            const auto requestJson = request.ToJson();
+            const auto& requestJson = request.ToJson();
             std::string jsonAsString = requestJson.toStyledString();
 
             std::unordered_map<std::string, std::string> headers;
             headers.emplace("X-EntityToken", request.authenticationContext == nullptr ? PlayFabSettings::entityToken : request.authenticationContext->entityToken);
 
-            auto reqContainer = std::unique_ptr<CallRequestContainer>(new CallRequestContainer(
+            std::unique_ptr<CallRequestContainer> reqContainer = std::unique_ptr<CallRequestContainer>(new CallRequestContainer(
```
ScottMunroMS: >std::unique_ptr<CallRequestContainer> [](start = 12, length = 37)

type is clearly on the right side. use of auto is fine here

source/code/source/playfab/QoS/PlayFabQoSApi.cpp
```diff
@@ -65,13 +65,13 @@ namespace PlayFab
         )
         {
             IPlayFabHttpPlugin& http = *PlayFabPluginManager::GetPlugin<IPlayFabHttpPlugin>(PlayFabPluginContract::PlayFab_Transport);
-            const auto requestJson = request.ToJson();
+            const auto& requestJson = request.ToJson();
```
ScottMunroMS: >auto [](start = 18, length = 4)

ToJson doesn't clearly state what the type is. recommend expanding to use the full type here

ScottMunroMS:
>             }

why copy the whole list and then do moves from the copy? you can avoid making copies of each of the Region strings and not incur any more ServerUrl copies if you just iterate over the original qosservers list

```
for (const QosServer& server : result.QosServers)
{
    api->regionMap[server.Region] = server.ServerUrl;
}
```

---
Refers to: source/code/source/playfab/QoS/PlayFabQoSApi.cpp:236 in b4b4959. [](commit_id = b4b495912a8523563b4af4a3d76c9199667e8688, deletion_comment = False)

source/code/source/playfab/QoS/XPlatSocket.cpp
```diff
@@ -107,7 +107,7 @@ namespace PlayFab
                 return GetLastErrorCode();
             }
 
-            auto inAddr = (struct in_addr*) he->h_addr_list[0];
+            in_addr* inAddr = (struct in_addr*) he->h_addr_list[0];
```
ScottMunroMS: > (struct in_addr*) [](start = 29, length = 18)

nit: you're using c++ so I recommend using a c++ style cast

`reinterpret_cast<in_addr*>(he->h_addr_list[0]);`

ScottMunroMS: >in_addr* [](start = 12, length = 8)

type is clearly on the right side so I think your use of auto was fine before

ScottMunroMS: >struct [](start = 31, length = 6)

nit: `struct` is unnecessary in c++

source/code/source/playfab/QoS/XPlatSocket.cpp
```diff
@@ -166,12 +166,12 @@ namespace PlayFab
             FD_CLR(0, &socketSet);
             FD_SET(s, &socketSet);
 
-            auto fd_max = s+1;
+            int fd_max = s+1;
```
ScottMunroMS: >s [](start = 25, length = 1)

`s` is a different type depending on platform (look at XPlatSocket.h)

instead of worrying about the type it may be simpler to just embed this in the select call

`select(s + 1, ...);`

source/code/source/playfab/QoS/XPlatSocket.cpp
```diff
@@ -166,12 +166,12 @@ namespace PlayFab
...
             int selectResult = select(fd_max, &socketSet, nullptr, nullptr, &timeOutVal);
 
             if (selectResult > 0)
             {
-                auto recvResult = recvfrom(s, buf, buflen, 0, (struct sockaddr *) &siOther, &slen);
+                size_t recvResult = recvfrom(s, buf, buflen, 0, (struct sockaddr *) &siOther, &slen);
```
ScottMunroMS: >size_t [](start = 16, length = 6)

on linux this is a ssize_t and on windows this is an int

in either case, size_t (which is unsigned) is incorrect and will never trigger the if-case on the line below

you already do a static_cast to int on line 179.

why not just make this value int from the get-go

templates/PlayFab_Api.cpp.ejs
```diff
@@ -38,13 +38,13 @@ namespace PlayFab
     {
 <%- getRequestActions("        ", apiCall, false) %>
         IPlayFabHttpPlugin& http = *PlayFabPluginManager::GetPlugin<IPlayFabHttpPlugin>(PlayFabPluginContract::PlayFab_Transport);
-        const auto requestJson = request.ToJson();
+        const auto& requestJson = request.ToJson();
```
ScottMunroMS: >auto [](start = 14, length = 4)

left comments elsewhere that I think should also be reflected in this template file

this use of auto should be an explicit type

the use of unique_ptr on 47 could be auto

internalPtr on 71 should be an explicit type

templates/PlayFab_InstanceApi.cpp.ejs
```diff
@@ -95,14 +95,14 @@ namespace PlayFab
     {
 <%- getRequestActions("        ", apiCall, true) %>
         IPlayFabHttpPlugin& http = *PlayFabPluginManager::GetPlugin<IPlayFabHttpPlugin>(PlayFabPluginContract::PlayFab_Transport);
-        const auto requestJson = request.ToJson();
+        const auto& requestJson = request.ToJson();
```
ScottMunroMS: similar concerns as elsewhere

explicit type here

auto instead of unique_ptr on 105

explicit type on 130 and 147

source/code/source/playfab/PlayFabCallRequestContainer.cpp
toddbellMSFT: This was showing up as a warning on the mac. I can pull it into its own PR.

source/code/source/playfab/QoS/XPlatSocket.cpp
toddbellMSFT: Int had issues when compiling on the Mac. ssize_t was the suggested fix over there. We may need to #ifdef here

source/code/source/playfab/PlayFabCallRequestContainer.cpp
ScottMunroMS: it can be its own PR if you'd like but I don't mind it being here.
I just think miscellaneous changes suckerfished onto other work should be explicitly called out/explained in the description if they are non-obvious. In this case I'd be interested to know what the warning is since there's nothing that stood out to me in the original block that would be substantially different (unless it was just a "deprecated API" warning or something)

source/code/source/playfab/QoS/XPlatSocket.cpp
ScottMunroMS: the documentation agrees that you should use `ssize_t` https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man2/recvfrom.2.html

but it's important to point out that what you used wasn't `ssize_t`. You used `size_t` (notice the missing 's' at the beginning for "signed")

I wouldn't recommend an ifdef.

You are already static-casting to int on the line that returns. Just static cast on the recvfrom line and save yourself the ifdef headache

source/code/source/playfab/PlayFabCallRequestContainer.cpp
toddbellMSFT: The warning stated Json::Reader was deprecated. The current jsoncpp github used the above as an example. I do like the unique_ptr change though. I'll be adding that in my next commit first.

ScottMunroMS: ah fair enough! then the change you've made is righteous, I just think it's worth calling out in the description so that others can figure out why the change was made later

source/code/source/playfab/PlayFabAndroidHttpPlugin.cpp
```diff
@@ -470,7 +471,7 @@ namespace PlayFab
 
         // Call SetBody
         {
-            auto requestUrl = GetUrl(requestTask);
+            const std::string& requestUrl = GetUrl(requestTask);
```
ScottMunroMS: >& [](start = 29, length = 1)

the other call to GetUrl was changed from a reference to value type but this one didn't change. Recommend doing another sweep

source/code/source/playfab/PlayFabAndroidHttpPlugin.cpp
ScottMunroMS: noticing that GetCallback didn't change in the latest iteration. intentional?

---
In reply to: [340920452](https://github.com/PlayFab/XPlatCoreTemplate/pull/33#discussion_r340920452) [](ancestors = 340920452,340919974)

source/code/source/playfab/QoS/PlayFabQoSApi.cpp
```diff
@@ -229,10 +229,9 @@ namespace PlayFab
             // Custom data received is a pointer to our api object
             PlayFabQoSApi* api = reinterpret_cast<PlayFabQoSApi*>(customData);
 
-            auto a = result.QosServers;
-            for (auto it = a.begin(); it != a.end(); ++it)
+            for(const QosServer& server :result.QosServers)
             {
-                api->regionMap[it->Region] = move(it->ServerUrl);
+                api->regionMap[server.Region] = move(server.ServerUrl);
```
ScottMunroMS: >move [](start = 48, length = 4)

I don't think you can (or want to) move this out of the server variable since it's a const reference.

You'll have to do a copy here but it'll still be N less copies than when you were copying the whole list before .

source/code/source/playfab/PlayFabEventBuffer.cpp
```diff
@@ -35,7 +35,7 @@ namespace PlayFab
 
         // First event is just a stub which is used only to maintain the initial consistency of m_head/m_tail.
         // It is immediately considered already consumed.
-        auto index = eventIndex->load(std::memory_order_relaxed);
+        auto index = static_cast<uint64_t>(eventIndex->load(std::memory_order_relaxed));
```
ScottMunroMS: > static_cast<uint64_t> [](start = 20, length = 22)

the static cast should be unnecessary. eventIndex is already a pointer to an atomic u64.

recommend removing the static cast and just writing the uint64_t type explicitly on the left-hand side

source/code/source/playfab/PlayFabEventBuffer.cpp
```diff
@@ -97,7 +97,7 @@ namespace PlayFab
 
         // create an event packet, set it to the tail->next and move the tail
 
-        const auto currentEventIndex = eventIndex->fetch_add(1, std::memory_order_relaxed);
+        const auto currentEventIndex = static_cast<const uint64_t>(eventIndex->fetch_add(1, std::memory_order_relaxed));
```
ScottMunroMS: >const [](start = 51, length = 5)

const is unnecessary here and similar comment about the static_cast being unnecessary as well

ScottMunroMS:
went ahead and approved. there are a couple of unresolved comments left but in case I don't have time to come back and check on them later you had fixed all the major things I noticed

PR #34, opened 2019-10-31
Some theoretically trivial changes to put Core & Switch more in-sync

pgilmorepf:
Also requestId parsing for Switch and Linux

### DIFF
--- a/source/code/source/playfab/PlayFabCurlHttpPlugin.cpp
+++ b/source/code/source/playfab/PlayFabCurlHttpPlugin.cpp
@@ -138,8 +138,8 @@ namespace PlayFab
             if (offset != std::string::npos)
             {
                 // Trim any whitespace
-                requestId.erase(0, offset);
                 requestId.erase(requestId.find_last_not_of(whitespace) + 1);
+                requestId.erase(0, offset);
                 // Save it
                 reqContainer.SetRequestId(requestId);
                 reqContainer.errorWrapper.RequestId = requestId;
--- a/source/code/source/playfab/PlayFabCurlHttpPlugin.cpp
+++ b/source/code/source/playfab/PlayFabCurlHttpPlugin.cpp
@@ -6,7 +6,12 @@
 
 #include <stdexcept>
 
-#define _countof(array) (sizeof(array) / sizeof(array[0]))
+#ifndef _countof
+template <typename _CountofType, size_t _SizeOfArray>
+char(*__countof_helper(_CountofType(&_Array)[_SizeOfArray]))[_SizeOfArray];
+
+#define _countof(_Array) (sizeof(*__countof_helper(_Array)) + 0)
+#endif
 
 namespace PlayFab
 {
@@ -118,6 +123,12 @@ namespace PlayFab
     size_t HeaderCallback(char* buffer, size_t size, size_t nitems, void* userdata)
     {
         CallRequestContainer& reqContainer = *static_cast<CallRequestContainer*>(userdata);
+        if (!reqContainer.errorWrapper.RequestId.empty())
+        {
+            // For now, we only care about the RequestId header
+            return nitems * size; // The return expected by curl for this callback
+        }
+
         // If this header-line is long enough, and the header starts with the key we expect
         if ((nitems > requestIdheaderKeyLen) && (strncasecmp(buffer, requestIdHeaderKey, requestIdheaderKeyLen) == 0))
         {
--- a/source/code/source/playfab/PlayFabCurlHttpPlugin.cpp
+++ b/source/code/source/playfab/PlayFabCurlHttpPlugin.cpp
@@ -116,13 +116,13 @@ namespace PlayFab
     constexpr char whitespace[] = "\t\n\v\f\r ";
     size_t HeaderCallback(char* buffer, size_t size, size_t nitems, void* userdata)
     {
-        size_t headerLen = _countof(requestIdHeaderKey) - 1;
+        size_t headerKeyLen = _countof(requestIdHeaderKey) - 1;
         CallRequestContainer& reqContainer = *static_cast<CallRequestContainer*>(userdata);
         // If this header-line is long enough, and the header starts with the key we expect
-        if ((nitems > headerLen) && (strncasecmp(buffer, requestIdHeaderKey, headerLen) == 0))
+        if ((nitems > headerKeyLen) && (strncasecmp(buffer, requestIdHeaderKey, headerKeyLen) == 0))
         {
             // The value is the requestId
-            std::string requestId = std::string(buffer + headerLen, nitems - headerLen);
+            std::string requestId = std::string(buffer + headerKeyLen, nitems - headerKeyLen);
             // Trim any whitespace
             requestId.erase(0, requestId.find_first_not_of(whitespace));
             requestId.erase(requestId.find_last_not_of(whitespace) + 1);
--- a/source/code/source/playfab/PlayFabCurlHttpPlugin.cpp
+++ b/source/code/source/playfab/PlayFabCurlHttpPlugin.cpp
@@ -113,22 +113,26 @@ namespace PlayFab
     }
 
     constexpr char requestIdHeaderKey[] = "X-RequestId:";
+    constexpr size_t requestIdheaderKeyLen = _countof(requestIdHeaderKey) - 1;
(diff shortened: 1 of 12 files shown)


pgilmorepf:
retest

source/code/source/playfab/PlayFabCurlHttpPlugin.cpp
```diff
@@ -110,6 +112,21 @@ namespace PlayFab
         } // UNLOCK httpRequestMutex
     }
 
+    constexpr char requestIdHeaderKey[] = "X-RequestId: ";
+    size_t HeaderCallback(char* buffer, size_t size, size_t nitems, void* userdata)
+    {
+        size_t headerLen = _countof(requestIdHeaderKey) - 1;
+        CallRequestContainer& reqContainer = *static_cast<CallRequestContainer*>(userdata);
+        // If this header starts with the key we expect
+        if (strncasecmp(buffer, requestIdHeaderKey, headerLen) == 0)
```
Vicodex: >strncasecmp(buffer, requestIdHeaderKey, headerLen) [](start = 12, length = 50)

Can't assume that the string passed in the buffer is always at least headerLen long (it can be shorter! and not necessarily end with 0!). I think you may also need to use size and nitems arguments here somehow to make it safe.

source/code/source/playfab/PlayFabCurlHttpPlugin.cpp
```diff
@@ -115,12 +115,13 @@ namespace PlayFab
     constexpr char requestIdHeaderKey[] = "X-RequestId: ";
```
jdeweyMSFT: The HTTP header spec allows for 0 or more linear whitespace after the colon, which are not considered part of the value. The value may also be followed by linear whitespace, which is not part of the value.
https://tools.ietf.org/html/rfc2616#section-4.2

source/code/source/playfab/PlayFabCurlHttpPlugin.cpp
pgilmorepf: Added a length test, which should prevent any potential buffer-overrun.

source/code/source/playfab/PlayFabCurlHttpPlugin.cpp
pgilmorepf: Since I'm already dropping the final requestId into a std::string, I can piggyback on that, add all the whitespace chars to that as well, and then use the std::string interface to remove them.

source/code/source/playfab/PlayFabCurlHttpPlugin.cpp
```diff
@@ -110,6 +112,27 @@ namespace PlayFab
         } // UNLOCK httpRequestMutex
     }
 
+    constexpr char requestIdHeaderKey[] = "X-RequestId:";
+    constexpr char whitespace[] = "\t\n\v\f\r ";
+    size_t HeaderCallback(char* buffer, size_t size, size_t nitems, void* userdata)
+    {
+        size_t headerLen = _countof(requestIdHeaderKey) - 1;
```
jdeweyMSFT: I would recommend calling this headerKeyLen, since it would be easy to misunderstand its meaning in later lines

source/code/source/playfab/PlayFabCurlHttpPlugin.cpp
```diff
@@ -110,6 +112,27 @@ namespace PlayFab
...
+        CallRequestContainer& reqContainer = *static_cast<CallRequestContainer*>(userdata);
+        // If this header-line is long enough, and the header starts with the key we expect
+        if ((nitems > headerLen) && (strncasecmp(buffer, requestIdHeaderKey, headerLen) == 0))
+        {
+            // The value is the requestId
+            std::string requestId = std::string(buffer + headerLen, nitems - headerLen);
+            // Trim any whitespace
+            requestId.erase(0, requestId.find_first_not_of(whitespace));
+            requestId.erase(requestId.find_last_not_of(whitespace) + 1);
```
jdeweyMSFT: if the value is only whitespace, the second erase will throw out of range.

Perhaps:
```cpp
auto offset = requestId.find_first_not_of(whitespace);
if (offset != std::string::npos)
{
    requestId.erase(0, offset);
    requestId.erase(requestId.find_last_not_of(whitespace) + 1);
}
```
The second erase is now only called if the string contains non-ws

source/code/source/playfab/PlayFabCurlHttpPlugin.cpp
```diff
@@ -6,6 +6,8 @@
 
 #include <stdexcept>
 
+#define _countof(array) (sizeof(array) / sizeof(array[0]))
```
KedarHMSFT: needs parentheses around the array part of array[0]

sizeof((array)[0])

have you considered using the countof helper we define in BumblelionArrayHelpers.h instead? It will prevent you from accidentally using this on a pointer type rather than an array type.

source/code/source/playfab/PlayFabCurlHttpPlugin.cpp
```diff
@@ -110,6 +112,27 @@ namespace PlayFab
         } // UNLOCK httpRequestMutex
     }
 
+    constexpr char requestIdHeaderKey[] = "X-RequestId:";
+    constexpr char whitespace[] = "\t\n\v\f\r ";
+    size_t HeaderCallback(char* buffer, size_t size, size_t nitems, void* userdata)
+    {
+        size_t headerKeyLen = _countof(requestIdHeaderKey) - 1;
```
KedarHMSFT: this could be a constexpr declared right under requestIdHeaderKey

source/code/source/playfab/PlayFabCurlHttpPlugin.cpp
pgilmorepf: This was copy-pasted from the official doc:
https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/countof-macro?view=vs-2019
I'm willing to use whatever seems best though.
I'll go grab the one in BumblelionArrayHelpers.h

source/code/source/playfab/PlayFabCurlHttpPlugin.cpp
pgilmorepf: _countof seems to conflict with constexpr

source/code/source/playfab/PlayFabCurlHttpPlugin.cpp
```diff
@@ -110,6 +112,27 @@ namespace PlayFab
         } // UNLOCK httpRequestMutex
     }
 
+    constexpr char requestIdHeaderKey[] = "X-RequestId:";
+    constexpr char whitespace[] = "\t\n\v\f\r ";
+    size_t HeaderCallback(char* buffer, size_t size, size_t nitems, void* userdata)
+    {
```
KedarHMSFT: I think you should add a check at the beginning of this callback that if the requestId is non-null then return early, to skip processing any other headers.

KedarHMSFT:
>         auto headers = reqContainer.GetRequestHeaders();

um, did this just make a *copy* of the map?
obligatory "auto is evil"

---
Refers to: source/code/source/playfab/PlayFabCurlHttpPlugin.cpp:154 in 18c3734. [](commit_id = 18c37346263c642e160d39684f3d0830adc073d7, deletion_comment = False)

source/code/source/playfab/PlayFabCurlHttpPlugin.cpp
```diff
@@ -110,6 +117,37 @@ namespace PlayFab
...
+        if ((nitems > requestIdheaderKeyLen) && (strncasecmp(buffer, requestIdHeaderKey, requestIdheaderKeyLen) == 0))
+        {
+            // The value is the requestId
+            std::string requestId = std::string(buffer + requestIdheaderKeyLen, nitems - requestIdheaderKeyLen);
+            auto offset = requestId.find_first_not_of(whitespace);
+            if (offset != std::string::npos)
+            {
+                // Trim any whitespace
+                requestId.erase(0, offset);
+                requestId.erase(requestId.find_last_not_of(whitespace) + 1);
```
KedarHMSFT: minor: if you reverse the order of these erase calls, then the erase at the beginning might have fewer bytes to copy when it shifts the contents of the string for the erase.

PR #35, opened 2019-11-01
Update PlayFabSpinLock.cpp

toddbellMSFT:
pulling in victors changes. acq_rel ensures we will be compare successfully with an acq_rel memory order. Therefore it's both an acquire and a release (which for comparing should work, according to documentation any memory order is possible, but for us, this would be the most flexible.


Vicodex:
I have an extensive explanation somewhere in the other project on why this is required :) 
It was also confirmed empirically with overnight testing.

PR #38, opened 2019-11-05
Marks most un-changed shared_ptr parameters to const refs

toddbellMSFT:
Most of these were being passed by value, therefore, copying the entire object whenever it was called. Passing the shared_ptr by const ref is still an unchanged standard. I'm leaving certain refs to shared_ptrs since we are writing to that object in the function, and we aren't passing by value, so we shouldn't suffer any copy pentalties.


source/code/source/playfab/PlayFabPluginManager.cpp
```diff
@@ -15,12 +15,12 @@ namespace PlayFab
     {
     }
 
-    void PlayFabPluginManager::SetPlugin(std::shared_ptr<IPlayFabPlugin> plugin, const PlayFabPluginContract contract, const std::string& instanceName)
+    void PlayFabPluginManager::SetPlugin(const std::shared_ptr<IPlayFabPlugin>& plugin, const PlayFabPluginContract contract, const std::string& instanceName)
```
pgilmorepf: agree

source/code/source/playfab/PlayFabPluginManager.cpp
```diff
@@ -15,12 +15,12 @@ namespace PlayFab
...
     {
         GetInstance().SetPluginInternal(std::move(plugin), contract, instanceName);
     }
 
-    void PlayFabPluginManager::SetPluginInstance(std::shared_ptr<IPlayFabPlugin> plugin, const PlayFabPluginContract contract, const std::string& instanceName)
+    void PlayFabPluginManager::SetPluginInstance(const std::shared_ptr<IPlayFabPlugin>& plugin, const PlayFabPluginContract contract, const std::string& instanceName)
```
pgilmorepf: agree

source/code/source/playfab/PlayFabPluginManager.cpp
```diff
@@ -55,7 +55,7 @@ namespace PlayFab
         }
     }
 
-    void PlayFabPluginManager::SetPluginInternal(std::shared_ptr<IPlayFabPlugin> plugin, const PlayFabPluginContract contract, const std::string& instanceName)
+    void PlayFabPluginManager::SetPluginInternal(const std::shared_ptr<IPlayFabPlugin>& plugin, const PlayFabPluginContract contract, const std::string& instanceName)
```
pgilmorepf: agree

source/code/source/playfab/PlayFabEventPipeline.cpp
```diff
@@ -335,7 +335,7 @@ namespace PlayFab
         }
     }
 
-    void PlayFabEventPipeline::CallbackRequest(std::shared_ptr<const IPlayFabEmitEventRequest> request, std::shared_ptr<const IPlayFabEmitEventResponse> response)
+    void PlayFabEventPipeline::CallbackRequest(const std::shared_ptr<const IPlayFabEmitEventRequest>& request, const std::shared_ptr<const IPlayFabEmitEventResponse>& response)
```
pgilmorepf: disagree

source/code/source/playfab/PlayFabEventPipeline.cpp
```diff
@@ -71,7 +71,7 @@ namespace PlayFab
         return this->settings;
     }
 
-    void PlayFabEventPipeline::IntakeEvent(std::shared_ptr<const IPlayFabEmitEventRequest> request)
+    void PlayFabEventPipeline::IntakeEvent(const std::shared_ptr<const IPlayFabEmitEventRequest>& request)
```
pgilmorepf: disagree

source/code/source/playfab/PlayFabEventPipeline.cpp
```diff
@@ -36,7 +36,7 @@ namespace PlayFab
     {
     }
 
-    PlayFabEventPipeline::PlayFabEventPipeline(std::shared_ptr<PlayFabEventPipelineSettings> settings) :
+    PlayFabEventPipeline::PlayFabEventPipeline(const std::shared_ptr<PlayFabEventPipelineSettings>& settings) :
```
pgilmorepf: agree
And, definitely need to remove the std:move(settings) below.

source/code/source/playfab/PlayFabEventBuffer.cpp
```diff
@@ -133,7 +133,7 @@ namespace PlayFab
         return EventConsumingResult::Success;
     }
 
-    PlayFabEventPacket* PlayFabEventBuffer::CreateEventPacket(uint8_t *location, const uint64_t index, std::shared_ptr<const IPlayFabEmitEventRequest> request)
+    PlayFabEventPacket* PlayFabEventBuffer::CreateEventPacket(uint8_t *location, const uint64_t index, const std::shared_ptr<const IPlayFabEmitEventRequest>& request)
```
pgilmorepf: We should never std::move a shared ptr. We should only move uniques.
This one should probably be a unique, and thus, not const&

source/code/source/playfab/PlayFabEventBuffer.cpp
```diff
@@ -57,7 +57,7 @@ namespace PlayFab
         }
     }
 
-    PlayFabEventBuffer::EventProducingResult PlayFabEventBuffer::TryPut(std::shared_ptr<const IPlayFabEmitEventRequest> request) // This must be thread-safe
+    PlayFabEventBuffer::EventProducingResult PlayFabEventBuffer::TryPut(const std::shared_ptr<const IPlayFabEmitEventRequest>& request) // This must be thread-safe
```
pgilmorepf: disagree

source/code/source/playfab/PlayFabEventApi.cpp
```diff
@@ -26,7 +26,7 @@ namespace PlayFab
         this->eventRouter->RouteEvent(eventRequest);
     }
 
-    void PlayFabEventAPI::EmitEvent(std::unique_ptr<const IPlayFabEvent> event, std::function<void(std::shared_ptr<const IPlayFabEvent>, std::shared_ptr<const IPlayFabEmitEventResponse>)> callback) const
+    void PlayFabEventAPI::EmitEvent(std::unique_ptr<const IPlayFabEvent> event, std::function<void(const std::shared_ptr<const IPlayFabEvent>&, const std::shared_ptr<const IPlayFabEmitEventResponse>&)> callback) const
```
pgilmorepf: disagree

source/code/source/playfab/QoS/PlayFabQoSApi.cpp
```diff
@@ -387,7 +387,7 @@ namespace PlayFab
         // Parameters : Configured socket to ping
         // Return : The ping result
         // Note that the function eat any exceptions thrown to it.
-        PingResult PlayFabQoSApi::GetQoSResultForRegion(shared_ptr<QoSSocket> socket)
+        PingResult PlayFabQoSApi::GetQoSResultForRegion(const shared_ptr<QoSSocket>& socket)
```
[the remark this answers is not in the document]
toddbellMSFT: Since the function below uses a non const labeled function, it does make sense to not make this const. That function does utilize unique member functions that may be "modified"

source/code/include/playfab/PlayFabCallRequestContainerBase.h
```diff
@@ -6,7 +6,7 @@
 namespace PlayFab
 {
     class CallRequestContainerBase;
-    typedef std::function<void(int, std::string, std::shared_ptr<CallRequestContainerBase>)> CallRequestContainerCallback;
+    typedef std::function<void(int, std::string, const std::shared_ptr<CallRequestContainerBase>&)> CallRequestContainerCallback;
```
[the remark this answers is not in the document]
toddbellMSFT: Since the callback is public facing AND we do want users to be granted the ability of transferring ownership. I can also agree with your disagreement here.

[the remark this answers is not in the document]
toddbellMSFT: Since this would be a public breaking change (users of any classes that derive this base class would need to change to a const ref param), I can get behind not taking this one.

source/code/source/playfab/PlayFabEventApi.cpp
toddbellMSFT: This derives from my reply above. This is a public facing callback function whose signature cannot change without breaking users.

source/code/source/playfab/PlayFabEventBuffer.cpp
toddbellMSFT: I tried changing this function signature to support a unique_ptr, but it causes some issues that would essentially lead to a larger refactor. So I'll accept the disagreement here too.

source/code/source/playfab/PlayFabEventPipeline.cpp
toddbellMSFT: Since "IntakeEvent" sounds like it would imply transfering ownership, sure, we can ignore this change too.

source/code/source/playfab/PlayFabEventPipeline.cpp
toddbellMSFT: Since this is based on a public facing callback function declaration that I stated in an earlier comment that we shouldn't change, then yeah we can revert this.

source/code/source/playfab/PlayFabEventRouter.cpp
```diff
@@ -17,7 +17,7 @@ namespace PlayFab
         this->pipelines.emplace(EventPipelineKey::PlayFabTelemetry, std::make_shared<PlayFabEventPipeline>(std::make_shared<PlayFabEventPipelineSettings>(PlayFabEventPipelineType::PlayFabTelemetry)));
     }
 
-    void PlayFabEventRouter::RouteEvent(std::shared_ptr<const IPlayFabEmitEventRequest> request) const
+    void PlayFabEventRouter::RouteEvent(const std::shared_ptr<const IPlayFabEmitEventRequest>& request) const
```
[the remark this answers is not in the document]
toddbellMSFT: The way this interacts with other functions that we have declared should be reverted. This one also makes sense to revert to conform to new standards.

source/code/source/playfab/PlayFabEventBuffer.cpp
toddbellMSFT: This gets a bit hairy when we swap to a unique ptr (this change trickles up to the PlayFabEventsApi::EmitEvent public function which will take me a bit to get a valid compile). The arguments passed to this function throughout this file is a std::move'd shared_ptr, which is a public function signature for TryPut (with that const &). SINCE Put is supposed to be taking ownership of the object, it makes sense TO make it something like a unique_ptr all the way through (but I would like to make sure it doesn't require the refactor we want to avoid).

source/code/include/playfab/PlayFabEventBuffer.h
```diff
@@ -16,7 +16,7 @@ namespace PlayFab
     class PlayFabEventPacket final
     {
     public:
-        PlayFabEventPacket(const uint64_t index, std::shared_ptr<const IPlayFabEmitEventRequest> request) :
+        PlayFabEventPacket(const uint64_t index, const std::shared_ptr<const IPlayFabEmitEventRequest>& request) :
```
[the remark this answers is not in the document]
toddbellMSFT: https://dev.azure.com/playfab/PlayFab/_workitems/edit/38671

PR #39, opened 2019-11-05
Instance signature and data-safety overhaul

pgilmorepf:
# Change description

* Several PlayFabSettings fields are moved to PlayFabSettings::staticPlayer or PlayFabSettings::staticSettings
* Some fields shuffled between PlayFabApiSettings and PlayFabAuthenticationContext based on appropriate-ness
* CallRequestContainer takes on more work, including referencing settings & context pointers and ValidateSettings
* New PlayFabException, which will replace std::runtime_error in several places, and will show up in more places in future commits
* Many settings & context member fields have been renamed to m_settings & m_context to avoid name conflicts
* Fixed standard tabbing issues in a lot of files
* Fixed multi-instance tests to not crash the program, if the calls fail.
* TestContext reports multiple failure outcomes slightly better.
* Instance API construction utilizes the more intuitive pattern established in C#
* Instance APIs will never read from, or write to the staticPlayer (This is the primary data leak)

# List of breaking changes:

High visibility breaking changes:

* PlayFabSettings::titleId is moved to PlayFabSettings::staticSettings->titleId
   * Severity: low, impacted customers: 100%, customer difficulty inflicted: easy
* PlayFabSettings::developerSecretKey is moved to PlayFabSettings::staticSettings->developerSecretKey
   * Severity: low, impacted customers: many, customer difficulty inflicted: easy

High impact, but low visibility breaking changes:

* PlayFabSettings::ValidateSettings() moved to CallRequestContainer
    * Severity: medium, impacted customers: ~none, customer difficulty inflicted: moderate
    * Defense: ValidateSettings is an internal use method which had no good reason to be in public PlayFabSettings
* Moved PlayFabApiSettings from CallRequestContainerBase to CallRequestContainer
    * Affects constructor, member methods and protected fields
    * Severity: medium, impacted customers: ~none, customer difficulty inflicted: moderate
    * Defense: It was a design violation for PlayFab specific objects to be in the base CRC
* Added PlayFabAuthenticationContext to CallRequestContainer
    * Affects constructor, member methods and protected fields
    * Severity: medium, impacted customers: ~none, customer difficulty inflicted: moderate
    * Defense: It's simply necessary for the instance data isolation. It's 100% required, and there's no good alternative.
* Added new PlayFabException throws from ApiInstance constructor under invalid conditions
    * Severity: medium, impacted customers: very low, customer difficulty inflicted: moderate
    * Defense: Any customer that triggers the new exception wasn't using instances correctly. If their previous instances worked, they were exploiting the data-leak bug that this commit will correct

Trivial breaking changes unlikely to affect customers

* PlayFabAuthenticationContext::developerSecretKey is moved to PlayFabApiSettings
   * Severity: low, impacted customers: few, customer difficulty inflicted: easy
* PlayFabSettings::verticalName removed
    * Severity: none, impacted customers: none, customer difficulty inflicted: none
* PlayFabApiSettings.verticalName removed
    * Severity: none, impacted customers: none, customer difficulty inflicted: none
* PlayFabSettings::requestParams moved to PlayFabApiSettings
    * Severity: very low, impacted customers: ~none, customer difficulty inflicted: none
* PlayFabSettings::useDevelopmentEnvironment removed
    * Severity: none, impacted customers: none, customer difficulty inflicted: none
* PlayFabSettings::developmentEnvironmentURL removed
    * Severity: none, impacted customers: none, customer difficulty inflicted: none
* PlayFabSettings::entityToken moved to PlayFabAuthenticationContext
    * Severity: low, impacted customers: very low, customer difficulty inflicted: easy
* PlayFabSettings::clientSessionTicket moved to PlayFabAuthenticationContext
    * Severity: low, impacted customers: very low, customer difficulty inflicted: easy
* PlayFabSettings::advertisingIdType moved to PlayFabAuthenticationContext
    * Severity: low, impacted customers: ~none, customer difficulty inflicted: easy
* PlayFabSettings::advertisingIdValue moved to PlayFabAuthenticationContext
    * Severity: low, impacted customers: ~none, customer difficulty inflicted: easy
* PlayFabSettings::GetUrl() redirected to same method in PlayFabApiSettings
    * Severity: low, impacted customers: ~none, customer difficulty inflicted: easy
* PlayFabSettings::GetUrl() method signature changed
    * Severity: low, impacted customers: ~none, customer difficulty inflicted: easy
* PlayFabApiSettings.advertisingIdType moved to PlayFabAuthenticationContext
    * Severity: low, impacted customers: ~none, customer difficulty inflicted: easy
* PlayFabApiSettings.advertisingIdValue moved to PlayFabAuthenticationContext
    * Severity: low, impacted customers: ~none, customer difficulty inflicted: easy
* PlayFabApiSettings.disableAdvertising moved to PlayFabAuthenticationContext
    * Severity: low, impacted customers: ~none, customer difficulty inflicted: easy
* Created PlayFabException, replaced several std::runtime_error instances with PlayFabException
    * Severity: none, impacted customers: none, customer difficulty inflicted: none
* Altered the print output of failing unit tests, for better debug-ability and readability
    * Severity: none, impacted customers: ~none, customer difficulty inflicted: ~none
* Fixed MultiStepClientLogin in API instances, such that it will work in circumstances where it would previously fail silently
    * Severity: none, impacted customers: ~none, customer difficulty inflicted: ~none

Because it's hard to see how template changes affect the SDK, I've made an example PR with the generated template here:
https://github.com/PlayFab/XPlatCppSdk/pull/30

### DIFF
--- a/source/code/include/playfab/PlayFabApiSettings.h
+++ b/source/code/include/playfab/PlayFabApiSettings.h
@@ -20,6 +20,12 @@ namespace PlayFab
 #endif
 
         PlayFabApiSettings();
+        PlayFabApiSettings(const PlayFabApiSettings& other) = delete;
+        PlayFabApiSettings(PlayFabApiSettings&& other) = delete;
+        PlayFabApiSettings& operator=(const PlayFabApiSettings& other) = delete;
+        PlayFabApiSettings& operator=(PlayFabApiSettings&& other) = delete;
+        ~PlayFabApiSettings() = default;
+
         std::string GetUrl(const std::string& urlPath) const;
     };
 }
--- a/source/code/include/playfab/PlayFabApiSettings.h
+++ b/source/code/include/playfab/PlayFabApiSettings.h
@@ -11,20 +11,15 @@ namespace PlayFab
     class PlayFabApiSettings
     {
     public:
-#ifndef DISABLE_PLAYFABCLIENT_API
-        std::string advertisingIdType; // Set this to the appropriate AD_TYPE_X constant (defined in PlayFabSettings)
-        std::string advertisingIdValue; // Set this to corresponding device value
+        static const std::map<std::string, std::string> requestGetParams;
 
-        // DisableAdvertising is provided for completeness, but changing it is not suggested
-        // Disabling this may prevent your advertising-related PlayFab marketplace partners from working correctly
-        bool disableAdvertising;
-#endif
-
-        std::string verticalName; // The name of a PlayFab service vertical
         std::string baseServiceHost; // The base for a PlayFab service host
         std::string titleId; // You must set this value for PlayFabSdk to work properly (found in the Game Manager for your title, at the PlayFab Website)
+#if defined(ENABLE_PLAYFABSERVER_API) || defined(ENABLE_PLAYFABADMIN_API)
+        std::string developerSecretKey; // Developer secret key. These keys can be used in development environments.
+#endif
 
         PlayFabApiSettings();
-        std::string GetUrl(const std::string& urlPath, const std::map<std::string, std::string>& getParams) const;
+        std::string GetUrl(const std::string& urlPath) const;
     };
 }
--- a/source/code/include/playfab/PlayFabAuthenticationContext.h
+++ b/source/code/include/playfab/PlayFabAuthenticationContext.h
@@ -11,16 +11,20 @@ namespace PlayFab
     {
     public:
 #ifndef DISABLE_PLAYFABCLIENT_API
+        // DisableAdvertising is provided for completeness, but changing it is not suggested
+        // Disabling this may prevent your advertising-related PlayFab marketplace partners from working correctly
+        bool disableAdvertising;
+        std::string playFabId; // Master_Player_Entity Id for the Player that logged in
         std::string clientSessionTicket; // Client session ticket that is used as an authentication token in many PlayFab API methods.
+        std::string advertisingIdType; // Set this to the appropriate AD_TYPE_X constant below
+        std::string advertisingIdValue; // Set this to corresponding device value
 #endif
-#ifndef DISABLE_PLAYFABENTITY_API
+        std::string entityId; // Entity Id for the active entity
+        std::string entityType; // Entity Type for the active entity
         std::string entityToken; // User's entity token. Entity tokens are required by all Entity API methods.
-#endif
-#if defined(ENABLE_PLAYFABSERVER_API) || defined(ENABLE_PLAYFABADMIN_API)
-        std::string developerSecretKey; // Developer secret key. These keys can be used in development environments.
-#endif
 
         PlayFabAuthenticationContext();
+        void HandlePlayFabLogin(const std::string& setPlayFabId, const std::string& setClientSessionTicket, const std::string& setEntityId, const std::string& setEntityType, const std::string& setEntityToken);
         void ForgetAllCredentials();
     };
 }
--- a/source/code/include/playfab/PlayFabAuthenticationContext.h
+++ b/source/code/include/playfab/PlayFabAuthenticationContext.h
@@ -24,6 +24,12 @@ namespace PlayFab
         std::string entityToken; // User's entity token. Entity tokens are required by all Entity API methods.
 
         PlayFabAuthenticationContext();
+        PlayFabAuthenticationContext(const PlayFabAuthenticationContext& other) = delete;
+        PlayFabAuthenticationContext(PlayFabAuthenticationContext&& other) = delete;
+        PlayFabAuthenticationContext& operator=(const PlayFabAuthenticationContext& other) = delete;
+        PlayFabAuthenticationContext& operator=(PlayFabAuthenticationContext&& other) = delete;
+        ~PlayFabAuthenticationContext() = default;
+
         void HandlePlayFabLogin(const std::string& setPlayFabId, const std::string& setClientSessionTicket, const std::string& setEntityId, const std::string& setEntityType, const std::string& setEntityToken);
         void ForgetAllCredentials();
     };
--- a/source/code/include/playfab/PlayFabAuthenticationContext.h
+++ b/source/code/include/playfab/PlayFabAuthenticationContext.h
@@ -30,7 +30,7 @@ namespace PlayFab
         PlayFabAuthenticationContext& operator=(PlayFabAuthenticationContext&& other) = delete;
         ~PlayFabAuthenticationContext() = default;
 
-        void HandlePlayFabLogin(const std::string& setPlayFabId, const std::string& setClientSessionTicket, const std::string& setEntityId, const std::string& setEntityType, const std::string& setEntityToken);
+        void HandlePlayFabLogin(const std::string& _playFabId, const std::string& _clientSessionTicket, const std::string& _entityId, const std::string& _entityType, const std::string& _entityToken);
         void ForgetAllCredentials();
     };
 }
--- a/source/code/include/playfab/PlayFabCallRequestContainer.h
+++ b/source/code/include/playfab/PlayFabCallRequestContainer.h
@@ -24,7 +24,7 @@ namespace PlayFab
         std::string GetFullUrl() const;
         std::shared_ptr<PlayFabApiSettings> GetApiSettings() const;
         std::shared_ptr<PlayFabAuthenticationContext> GetContext() const;
-        bool ValidateSettings();
+        bool HandleInvalidSettings();
 
         // TODO: clean up these public variables with setters/getters when you have the chance.
 
--- a/source/code/include/playfab/PlayFabCallRequestContainer.h
+++ b/source/code/include/playfab/PlayFabCallRequestContainer.h
@@ -1,5 +1,6 @@
 #pragma once
 
+#include <playfab/PlayFabApiSettings.h>
 #include <playfab/PlayFabError.h>
 #include <playfab/PlayFabCallRequestContainerBase.h>
 
@@ -21,7 +22,9 @@ namespace PlayFab
 
         virtual ~CallRequestContainer() override;
         std::string GetFullUrl() const;
-        virtual bool ValidateSettings() override;
+        std::shared_ptr<PlayFabApiSettings> GetApiSettings() const;
+        std::shared_ptr<PlayFabAuthenticationContext> GetContext() const;
+        bool ValidateSettings();
 
         // TODO: clean up these public variables with setters/getters when you have the chance.
 
@@ -31,5 +34,7 @@ namespace PlayFab
         PlayFabError errorWrapper;
         std::shared_ptr<void> successCallback;
         ErrorCallback errorCallback;
+        std::shared_ptr<PlayFabApiSettings> m_settings;
+        std::shared_ptr<PlayFabAuthenticationContext> m_context;
     };
 }
--- a/source/code/include/playfab/PlayFabCallRequestContainer.h
+++ b/source/code/include/playfab/PlayFabCallRequestContainer.h
@@ -6,7 +6,7 @@
 namespace PlayFab
 {
     /// <summary>
-    /// Internal PlayFabHttp container for each api call
+    /// Internal PlayFabHttp container for each API call
     /// </summary>
     class CallRequestContainer : public CallRequestContainerBase
     {
@@ -15,11 +15,13 @@ namespace PlayFab
             const std::unordered_map<std::string, std::string>& headers,
             const std::string& requestBody,
             CallRequestContainerCallback callback,
-            void* customData = nullptr,
-            std::shared_ptr<PlayFabApiSettings> apiSettings = nullptr);
+            std::shared_ptr<PlayFabApiSettings> apiSettings,
+            std::shared_ptr<PlayFabAuthenticationContext> context,
+            void* customData);
 
         virtual ~CallRequestContainer() override;
         std::string GetFullUrl() const;
+        virtual bool ValidateSettings() override;
 
         // TODO: clean up these public variables with setters/getters when you have the chance.
 
--- a/source/code/include/playfab/PlayFabCallRequestContainerBase.h
+++ b/source/code/include/playfab/PlayFabCallRequestContainerBase.h
@@ -1,6 +1,5 @@
 #pragma once
 
-#include <playfab/PlayFabApiSettings.h>
 #include <unordered_map>
 #include <memory>
 
@@ -27,9 +26,7 @@ namespace PlayFab
             const std::unordered_map<std::string, std::string>& headers,
             std::string requestBody,
             CallRequestContainerCallback callback,
-            std::shared_ptr<PlayFabApiSettings> apiSettings,
-            std::shared_ptr<PlayFabAuthenticationContext> context,
-            void* customData);
+            void* customData = nullptr);
 
         virtual ~CallRequestContainerBase() = default;
 
@@ -38,8 +35,6 @@ namespace PlayFab
         std::string GetRequestId() const;
         void SetRequestId(const std::string& newRequestId);
         std::string GetRequestBody() const;
-        std::shared_ptr<PlayFabApiSettings> GetApiSettings() const;
-        std::shared_ptr<PlayFabAuthenticationContext> GetContext() const;
 
         /// <summary>
         /// This function is meant to handle logic of calling the error callback or success
@@ -47,15 +42,12 @@ namespace PlayFab
         CallRequestContainerCallback GetCallback() const;
 
         void* GetCustomData() const;
-        virtual bool ValidateSettings() = 0;
 
     protected:
         std::string url;
         std::unordered_map<std::string, std::string> requestHeaders;
         std::string requestBody;
         std::string requestId;
-        std::shared_ptr<PlayFabApiSettings> m_settings;
-        std::shared_ptr<PlayFabAuthenticationContext> m_context;
         CallRequestContainerCallback callback;
 
         // I never own this, I can never destroy it
--- a/source/code/include/playfab/PlayFabCallRequestContainerBase.h
+++ b/source/code/include/playfab/PlayFabCallRequestContainerBase.h
@@ -27,8 +27,9 @@ namespace PlayFab
             const std::unordered_map<std::string, std::string>& headers,
             std::string requestBody,
             CallRequestContainerCallback callback,
-            void* customData = nullptr,
-            std::shared_ptr<PlayFabApiSettings> apiSettings = nullptr);
+            std::shared_ptr<PlayFabApiSettings> apiSettings,
+            std::shared_ptr<PlayFabAuthenticationContext> context,
+            void* customData);
 
         virtual ~CallRequestContainerBase() = default;
 
@@ -38,6 +39,7 @@ namespace PlayFab
         void SetRequestId(const std::string& newRequestId);
         std::string GetRequestBody() const;
         std::shared_ptr<PlayFabApiSettings> GetApiSettings() const;
+        std::shared_ptr<PlayFabAuthenticationContext> GetContext() const;
 
         /// <summary>
         /// This function is meant to handle logic of calling the error callback or success
@@ -45,13 +47,15 @@ namespace PlayFab
         CallRequestContainerCallback GetCallback() const;
 
         void* GetCustomData() const;
+        virtual bool ValidateSettings() = 0;
 
     protected:
         std::string url;
         std::unordered_map<std::string, std::string> requestHeaders;
         std::string requestBody;
         std::string requestId;
-        std::shared_ptr<PlayFabApiSettings> apiSettings;
+        std::shared_ptr<PlayFabApiSettings> m_settings;
+        std::shared_ptr<PlayFabAuthenticationContext> m_context;
         CallRequestContainerCallback callback;
 
         // I never own this, I can never destroy it
--- a/source/code/include/playfab/PlayFabError.h.ejs
+++ b/source/code/include/playfab/PlayFabError.h.ejs
@@ -42,4 +42,29 @@ namespace PlayFab
 
     typedef std::function<void(const PlayFabError& error, void* customData)> ErrorCallback;
     typedef std::function<void(std::exception exception)> ExceptionCallback;
+
+    enum class PlayFabExceptionCode
+    {
+        AuthContextRequired,
+        DeveloperKeyNotSet,
+        EntityTokenNotSet,
+        NotLoggedIn,
+        PluginAmbiguity,
+        PluginNotFound,
+        ThreadMisuse,
+        TitleNotSet,
+    };
+
+    class PlayFabException : public std::runtime_error
+    {
+    public:
+        PlayFabException() = delete;
+        PlayFabException(const PlayFabException& source) = default;
+        PlayFabException(PlayFabException&&) = default;
+        PlayFabException& operator=(const PlayFabException& source) = default;
+        PlayFabException& operator=(PlayFabException&& other) = default;
+
+        PlayFabExceptionCode Code;
+        PlayFabException(PlayFabExceptionCode code, const char* const message);
+    };
 }
--- a/source/code/include/playfab/PlayFabError.h.ejs
+++ b/source/code/include/playfab/PlayFabError.h.ejs
@@ -59,10 +59,10 @@ namespace PlayFab
     {
     public:
         PlayFabException() = delete;
-        PlayFabException(const PlayFabException& source) = default;
+        PlayFabException(const PlayFabException& source) = delete;
         PlayFabException(PlayFabException&&) = default;
-        PlayFabException& operator=(const PlayFabException& source) = default;
-        PlayFabException& operator=(PlayFabException&& other) = default;
+        PlayFabException& operator=(const PlayFabException& source) = delete;
+        PlayFabException& operator=(PlayFabException&& other) = delete;
 
         PlayFabExceptionCode Code;
         PlayFabException(PlayFabExceptionCode code, const char* const message);
--- a/source/code/include/playfab/PlayFabEventPipeline.h
+++ b/source/code/include/playfab/PlayFabEventPipeline.h
@@ -88,7 +88,8 @@ namespace PlayFab
         std::vector<std::shared_ptr<const IPlayFabEmitEventRequest>> batch;
 
     private:
-        std::shared_ptr<PlayFabEventPipelineSettings> settings;
+        std::shared_ptr<PlayFabEventPipelineSettings> m_settings;
+        std::shared_ptr<PlayFabAuthenticationContext> m_context;
         PlayFabEventBuffer buffer;
         std::thread workerThread;
         std::atomic<bool> isWorkerThreadRunning;
--- a/source/code/include/playfab/PlayFabEventPipeline.h
+++ b/source/code/include/playfab/PlayFabEventPipeline.h
@@ -89,7 +89,6 @@ namespace PlayFab
 
     private:
         std::shared_ptr<PlayFabEventPipelineSettings> settings;
-        std::shared_ptr<PlayFabAuthenticationContext> context;
         PlayFabEventBuffer buffer;
         std::thread workerThread;
         std::atomic<bool> isWorkerThreadRunning;
--- a/source/code/include/playfab/PlayFabEventPipeline.h
+++ b/source/code/include/playfab/PlayFabEventPipeline.h
@@ -88,8 +88,8 @@ namespace PlayFab
         std::vector<std::shared_ptr<const IPlayFabEmitEventRequest>> batch;
 
     private:
-        std::shared_ptr<PlayFabEventPipelineSettings> m_settings;
-        std::shared_ptr<PlayFabAuthenticationContext> m_context;
+        std::shared_ptr<PlayFabEventPipelineSettings> settings;
+        std::shared_ptr<PlayFabAuthenticationContext> context;
         PlayFabEventBuffer buffer;
         std::thread workerThread;
         std::atomic<bool> isWorkerThreadRunning;
--- a/source/code/include/playfab/PlayFabSettings.h
+++ b/source/code/include/playfab/PlayFabSettings.h
@@ -15,39 +15,22 @@ namespace PlayFab
         static const std::string sdkVersion;
         static const std::string buildIdentifier;
         static const std::string versionString;
-        static const std::string verticalName;
 
-        static const std::map<std::string, std::string> requestGetParams;
-
-        static bool useDevelopmentEnvironment;
-        static std::string developmentEnvironmentURL;
         static std::string productionEnvironmentURL;
-        static std::string titleId; // You must set this value for PlayFabSdk to work properly (Found in the Game Manager for your title, at the PlayFab Website)
         static ErrorCallback globalErrorHandler;
 
-        // Control whether all callbacks are threaded or whether the user manually controlls callback timing from their main-thread
+        static std::shared_ptr<PlayFabApiSettings> staticSettings;
+        static std::shared_ptr<PlayFabAuthenticationContext> staticPlayer;
+
+        // Control whether all callbacks are threaded or whether the user manually controls callback timing from their main-thread
         static bool threadedCallbacks;
 
-        static std::string entityToken; // This is set by entity GetEntityToken method, and is required by all other Entity API methods
-#if defined(ENABLE_PLAYFABSERVER_API) || defined(ENABLE_PLAYFABADMIN_API)
-        static std::string developerSecretKey; // You must set this value for PlayFabSdk to work properly (Found in the Game Manager for your title, at the PlayFab Website)
-#endif
 #ifndef DISABLE_PLAYFABCLIENT_API
-        static std::string clientSessionTicket; // This is set by any Client Login method, and is required for all other Client API methods
-        static std::string advertisingIdType; // Set this to the appropriate AD_TYPE_X constant below
-        static std::string advertisingIdValue; // Set this to corresponding device value
-
-        // DisableAdvertising is provided for completeness, but changing it is not suggested
-        // Disabling this may prevent your advertising-related PlayFab marketplace partners from working correctly
-        static bool disableAdvertising;
         static const std::string AD_TYPE_IDFA;
         static const std::string AD_TYPE_ANDROID_ID;
+#endif
 
         static void ForgetAllCredentials();
-
-        static std::string GetUrl(const std::string& urlPath, const std::map<std::string, std::string>& getParams);
-#endif
-        static bool ValidateSettings(const std::shared_ptr<PlayFabAuthenticationContext> authenticationContext, const std::shared_ptr<PlayFabApiSettings> apiSettings, CallRequestContainer& container);
     private:
         PlayFabSettings(); // Private constructor, static class should never have an instance
         PlayFabSettings(const PlayFabSettings& other); // Private copy-constructor, static class should never have an instance
--- a/source/code/include/playfab/PlayFabSettings.h
+++ b/source/code/include/playfab/PlayFabSettings.h
@@ -16,14 +16,16 @@ namespace PlayFab
         static const std::string buildIdentifier;
         static const std::string versionString;
 
-        static std::string productionEnvironmentURL;
-        static ErrorCallback globalErrorHandler;
-
-        static std::shared_ptr<PlayFabApiSettings> staticSettings;
-        static std::shared_ptr<PlayFabAuthenticationContext> staticPlayer;
-
         // Control whether all callbacks are threaded or whether the user manually controls callback timing from their main-thread
         static bool threadedCallbacks;
+        // Used to override the PlayFab endpoint url - Not typical
+        static std::string productionEnvironmentURL;
+        // Used to receive a callback for every failed PlayFab API call - Parallel to the individual error callbacks
+        static ErrorCallback globalErrorHandler;
+
+        // The pointers to these objects should be const as they should always be fixed, but the contents are still mutable
+        static const std::shared_ptr<PlayFabApiSettings> staticSettings;
+        static const std::shared_ptr<PlayFabAuthenticationContext> staticPlayer;
 
 #ifndef DISABLE_PLAYFABCLIENT_API
         static const std::string AD_TYPE_IDFA;
--- a/source/code/source/playfab/PlayFabAndroidHttpPlugin.cpp
+++ b/source/code/source/playfab/PlayFabAndroidHttpPlugin.cpp
@@ -605,7 +605,7 @@ namespace PlayFab
 
     std::string PlayFabAndroidHttpPlugin::GetUrl(const RequestTask& requestTask) const
     {
-        return PlayFabSettings::GetUrl(requestTask.GetRequestContainerUrl(), PlayFabSettings::staticSettings->requestGetParams);
+        return requestTask.GetRequestContainerUrl();
     }
 
     void PlayFabAndroidHttpPlugin::SetPredefinedHeaders(const RequestTask& requestTask)
--- a/source/code/source/playfab/PlayFabAndroidHttpPlugin.cpp
+++ b/source/code/source/playfab/PlayFabAndroidHttpPlugin.cpp
@@ -247,6 +247,11 @@ namespace PlayFab
 
     void PlayFabAndroidHttpPlugin::MakePostRequest(std::unique_ptr<CallRequestContainerBase> requestContainer)
     {
+        if (!requestContainer->ValidateSettings())
+        {
+            return;
+        }
+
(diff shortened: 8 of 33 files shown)


source/code/include/playfab/PlayFabCallRequestContainer.h
```diff
@@ -15,11 +15,13 @@ namespace PlayFab
             const std::unordered_map<std::string, std::string>& headers,
             const std::string& requestBody,
             CallRequestContainerCallback callback,
-            void* customData = nullptr,
-            std::shared_ptr<PlayFabApiSettings> apiSettings = nullptr);
+            std::shared_ptr<PlayFabApiSettings> apiSettings,
```
toddbellMSFT: nit: any particular reason these don't need to default to nullptr anymore?

source/code/source/playfab/PlayFabApiSettings.cpp
```diff
@@ -5,39 +5,33 @@
...
+        {
+            titleId = PlayFabSettings::staticSettings->titleId;
+        }
     }
 
-    std::string PlayFabApiSettings::GetUrl(const std::string& urlPath, const std::map<std::string, std::string>& getParams) const
+    std::string PlayFabApiSettings::GetUrl(const std::string& urlPath) const
     {
         std::string fullUrl;
         fullUrl.reserve(1000);
```
toddbellMSFT: side note: With the removal of getParams here in favor of the object created above, should we consider a finder reserve number for the fullUrl instead of an arbitrary 1000?

source/test/TestApp/PlayFabTestMultiUserInstance.cpp
```diff
@@ -23,21 +23,21 @@ namespace PlayFabUnit
 
         const auto& user1ProfileSuccess = std::bind(&PlayFabTestMultiUserInstance::MultiUserProfile1Success, this, std::placeholders::_1, std::placeholders::_2);
         const auto& user1ProfileFailure = std::bind(&PlayFabTestMultiUserInstance::MultiUserProfile1Failure, this, std::placeholders::_1, std::placeholders::_2);
-        (*multiUser1ClientApi)->GetPlayerProfile(profileRequest, user1ProfileSuccess, user1ProfileFailure, customData);
+        multiUser1ClientApi->GetPlayerProfile(profileRequest, user1ProfileSuccess, user1ProfileFailure, customData);
     }
-    void PlayFabTestMultiUserInstance::MultiUserLogin1Failure(const PlayFabError& error, void* customData)
+    void PlayFabTestMultiUserInstance::MultiUserLogin1Failure(const PlayFabError& error, void* /*customData*/)
```
toddbellMSFT: Awesome :) I ran into this warning today about unused variables, I'll follow the pattern you have here.

templates/PlayFab_InstanceApi.cpp.ejs
```diff
@@ -140,39 +158,21 @@ namespace PlayFab
     // Private PlayFabClientInstanceAPI specific
     bool PlayFabClientInstanceAPI::IsClientLoggedIn()
     {
-        return !this->GetOrCreateAuthenticationContext()->clientSessionTicket.empty();
+        return !this->m_context->clientSessionTicket.empty();
     }
 
-    void PlayFabClientInstanceAPI::MultiStepClientLogin(bool needsAttribution)
+    void PlayFabClientInstanceAPI::MultiStepClientLogin(std::shared_ptr<PlayFabAuthenticationContext> context, bool needsAttribution)
```
toddbellMSFT: nit: can this shared_ptr be passed by const & instead of copied by value?

templates/PlayFab_InstanceApi.cpp.ejs
```diff
@@ -140,39 +158,21 @@ namespace PlayFab
...
-                if (apiSettings->advertisingIdType == PlayFabSettings::AD_TYPE_IDFA)
-                    request.Idfa = apiSettings->advertisingIdValue;
-                else if (apiSettings->advertisingIdType == PlayFabSettings::AD_TYPE_ANDROID_ID)
-                    request.Adid = apiSettings->advertisingIdValue;
-                else
-                    return;
-                AttributeInstall(request, nullptr, nullptr);
-            }
+            AttributeInstallRequest request;
+            if (context->advertisingIdType == PlayFabSettings::AD_TYPE_IDFA)
```
debhaldarMS: please use braces - here and other places .

templates/PlayFab_InstanceApi.cpp.ejs
```diff
@@ -140,39 +158,21 @@ namespace PlayFab
...
+                request.Idfa = context->advertisingIdValue;
+            else if (context->advertisingIdType == PlayFabSettings::AD_TYPE_ANDROID_ID)
+                request.Adid = context->advertisingIdValue;
+            else
+                return;
```
debhaldarMS: redundant else and return ?

source/code/include/playfab/PlayFabSettings.h
```diff
@@ -15,39 +15,22 @@ namespace PlayFab
         static const std::string sdkVersion;
         static const std::string buildIdentifier;
         static const std::string versionString;
-        static const std::string verticalName;
```
debhaldarMS: > verticalName; [](start = 32, length = 14)

this is an api breaking change - we need to think if we can avoid that ?

source/code/source/playfab/PlayFabApiSettings.cpp
```diff
@@ -5,39 +5,33 @@
 
 namespace PlayFab
 {
+    const std::map<std::string, std::string> PlayFabApiSettings::requestGetParams = {
```
KedarHMSFT: can this whole map be deleted and instead use a hardcoded string literal that gets appended on the URL?

source/code/include/playfab/PlayFabCallRequestContainer.h
pgilmorepf: 1. I changed the order of the parameters
2. I wanted to ensure that every place where this is made, it's made with deliberate intent

What I mean with 2 is that we still pass nullptr sometimes, but we pass it deliberately, not by accident.

source/code/source/playfab/PlayFabApiSettings.cpp
pgilmorepf: I'm not opposed to a new feature task to improve that 1000.
But, I am opposed to doing it in this checkin.

templates/PlayFab_InstanceApi.cpp.ejs
pgilmorepf: Let's do it as part of your checkin for this purpose.
I think you're right though, it is a safe candidate here.

templates/PlayFab_InstanceApi.cpp.ejs
pgilmorepf: False. That else condition is the fail-out.
If we don't have data to send, we won't call AttributeInstall on the next line.

source/code/include/playfab/PlayFabSettings.h
pgilmorepf: verticalName:

1. Never worked
2. Design-wise is totally wrong
2. Will break everything if you set it
3. Won't be used in a future scenario where we do the right thing

source/code/source/playfab/PlayFabApiSettings.cpp
pgilmorepf: Not trivially. The requestParams are already designed using this pattern, and they're intended to be mutable: We have ours, customers could have theirs, and they need to be properly merged.
I'm willing to consider a future task to improve the performance and keep the mutability, but improve the performance.
Oh... const... bah, that const may not be correct...
I'll file the task.

pgilmorepf: (Task filed)

templates/PlayFab_InstanceApi.cpp.ejs
toddbellMSFT: [https://github.com/PlayFab/XPlatCoreTemplate/pull/40](https://github.com/PlayFab/XPlatCoreTemplate/pull/40) I will fix this issue in a later PR.

---
In reply to: [342834511](https://github.com/PlayFab/XPlatCoreTemplate/pull/39#discussion_r342834511) [](ancestors = 342834511)

pgilmorepf: Appropriate braces added to everything I touched.

---
In reply to: [343240742](https://github.com/PlayFab/XPlatCoreTemplate/pull/39#discussion_r343240742) [](ancestors = 343240742,342834511)

source/test/TestApp/PlayFabTestMultiUserInstance.cpp
ScottMunroMS: FWIW, another option we use in bumblelion is to just mark the parameter unused in the function body
```
UNREFERENCED_PARAMETER(customData); // this may be a windows macro
(void)customData; // equivalent to the macro above
```

another option is to just leave the name unspecified
`MultiUserLogin1Failure(const PlayFabError& error, void*);`

templates/PlayFab_InstanceApi.cpp.ejs
ScottMunroMS: If you are passing shared_ptr's by const value you should evaluate whether you need to be passing a shared_ptr at all. Does the function being called take shared ownership of the resource? then you should just pass by std::shared_ptr and std::move the parameter whenever ownership needs to be transferred in the function body. If it doesn't need to provide shared_ownership than just pass a raw pointer/reference `Func(const PlayFabAuthenticationContext& context, ...)`

source/code/include/playfab/PlayFabSettings.h
ScottMunroMS: It surprises me to hear you say these things about this publicly exposed data member

1. If it never worked, it's never been tested or the tests were broken
2. If its design is totally wrong then its design was never exercised in any sort of proof-of-concept playground (likely an integration test which is likely missing)
3. If it breaks everything when set, it's double definitely never been 

I'm not sure what the history of this code/feature is, but it seems like something is in crucial need of reworking if a wrongly designed, broken feature made it into the public SDK.

source/code/include/playfab/PlayFabAuthenticationContext.h
```diff
@@ -11,16 +11,20 @@ namespace PlayFab
...
+        std::string entityId; // Entity Id for the active entity
+        std::string entityType; // Entity Type for the active entity
         std::string entityToken; // User's entity token. Entity tokens are required by all Entity API methods.
-#endif
-#if defined(ENABLE_PLAYFABSERVER_API) || defined(ENABLE_PLAYFABADMIN_API)
-        std::string developerSecretKey; // Developer secret key. These keys can be used in development environments.
-#endif
 
         PlayFabAuthenticationContext();
+        void HandlePlayFabLogin(const std::string& setPlayFabId, const std::string& setClientSessionTicket, const std::string& setEntityId, const std::string& setEntityType, const std::string& setEntityToken);
```
debhaldarMS: >set [](start = 51, length = 3)

Why are these named as with "set" prefix ? Is that a convention we're following in other parts of the code ?

source/code/include/playfab/PlayFabApiSettings.h
```diff
@@ -11,20 +11,15 @@ namespace PlayFab
     class PlayFabApiSettings
```
debhaldarMS: >PlayFabApiSettings [](start = 10, length = 18)

Do you anticipate this class to be copy or move constructed / assigned?

debhaldarMS:
> namespace PlayFab

Do you anticipate this class to be copy or move constructed / assigned?

---
Refers to: source/code/include/playfab/PlayFabAuthenticationContext.h:5 in 5fd376d. [](commit_id = 5fd376dd148fe94f311c25bcd19d895397ae3a30, deletion_comment = False)

source/code/include/playfab/PlayFabCallRequestContainerBase.h
```diff
@@ -38,20 +39,23 @@ namespace PlayFab
...
         std::shared_ptr<PlayFabApiSettings> GetApiSettings() const;
+        std::shared_ptr<PlayFabAuthenticationContext> GetContext() const;
 
         /// <summary>
         /// This function is meant to handle logic of calling the error callback or success
         /// </summary>
         CallRequestContainerCallback GetCallback() const;
 
         void* GetCustomData() const;
+        virtual bool ValidateSettings() = 0;
```
debhaldarMS: >virtual bool ValidateSettings() = 0; [](start = 8, length = 36)

Please do not add pure virtual methods to already released APIs. This is a bad practice in general. For an explanation - check out #20 here: [https://www.acodersjourney.com/top-25-cplusplus-api-design-mistakes-and-how-to-avoid-them/](https://www.acodersjourney.com/top-25-cplusplus-api-design-mistakes-and-how-to-avoid-them/)

source/code/include/playfab/PlayFabCallRequestContainerBase.h
```diff
@@ -38,20 +39,23 @@ namespace PlayFab
...
 
     protected:
         std::string url;
         std::unordered_map<std::string, std::string> requestHeaders;
         std::string requestBody;
         std::string requestId;
-        std::shared_ptr<PlayFabApiSettings> apiSettings;
+        std::shared_ptr<PlayFabApiSettings> m_settings;
```
debhaldarMS: >m_ [](start = 44, length = 2)

I thought we were not using the m_ as a convention.

source/code/include/playfab/PlayFabError.h.ejs
```diff
@@ -42,4 +42,29 @@ namespace PlayFab
...
+        DeveloperKeyNotSet,
+        EntityTokenNotSet,
+        NotLoggedIn,
+        PluginAmbiguity,
+        PluginNotFound,
+        ThreadMisuse,
+        TitleNotSet,
+    };
+
+    class PlayFabException : public std::runtime_error
```
debhaldarMS: >runtime_error [](start = 41, length = 13)

why are we not deriving from std::exception ? will catch(...) still handle this ?

source/code/include/playfab/PlayFabEventPipeline.h
```diff
@@ -88,7 +88,8 @@ namespace PlayFab
         std::vector<std::shared_ptr<const IPlayFabEmitEventRequest>> batch;
 
     private:
-        std::shared_ptr<PlayFabEventPipelineSettings> settings;
+        std::shared_ptr<PlayFabEventPipelineSettings> m_settings;
```
debhaldarMS: > m_ [](start = 53, length = 3)

remove m_

source/code/source/playfab/PlayFabAuthenticationContext.cpp
```diff
@@ -7,27 +7,28 @@ namespace PlayFab
...
-}
\ No newline at end of file
+
+    void PlayFabAuthenticationContext::HandlePlayFabLogin(const std::string& setPlayFabId, const std::string& setClientSessionTicket, const std::string& setEntityId, const std::string& setEntityType, const std::string& setEntityToken)
+    {
+        if (!setPlayFabId.empty()) playFabId = setPlayFabId;
+        if (!setClientSessionTicket.empty()) clientSessionTicket = setClientSessionTicket;
+        if (!setEntityId.empty()) entityId = setEntityId;
+        if (!setEntityType.empty()) entityType = setEntityType;
+        if (!setEntityToken.empty()) entityToken = setEntityToken;
```
debhaldarMS: seperate lines with braces

source/code/source/playfab/PlayFabAuthenticationContext.cpp
```diff
@@ -7,27 +7,28 @@ namespace PlayFab
...
         entityToken.clear();
-#endif
-#if defined(ENABLE_PLAYFABSERVER_API) || defined(ENABLE_PLAYFABADMIN_API)
-        developerSecretKey.clear();
-#endif
     }
-}
\ No newline at end of file
+
+    void PlayFabAuthenticationContext::HandlePlayFabLogin(const std::string& setPlayFabId, const std::string& setClientSessionTicket, const std::string& setEntityId, const std::string& setEntityType, const std::string& setEntityToken)
```
debhaldarMS: >(const std::string& setPlayFabId, const std::string& setClientSessionTicket, const std::string& setEntityId, const std::string& setEntityType, const std::string& setEntityToken) [](start = 57, length = 177)

please consider breaking this down in multiple lines

source/code/include/playfab/PlayFabSettings.h
```diff
@@ -15,39 +15,22 @@ namespace PlayFab
...
-        static const std::map<std::string, std::string> requestGetParams;
-
-        static bool useDevelopmentEnvironment;
-        static std::string developmentEnvironmentURL;
         static std::string productionEnvironmentURL;
-        static std::string titleId; // You must set this value for PlayFabSdk to work properly (Found in the Game Manager for your title, at the PlayFab Website)
         static ErrorCallback globalErrorHandler;
 
-        // Control whether all callbacks are threaded or whether the user manually controlls callback timing from their main-thread
+        static std::shared_ptr<PlayFabApiSettings> staticSettings;
```
debhaldarMS: >shared_ptr [](start = 20, length = 10)

Any reason why this one is not unique pointer ? Will this be modified in any way ?

source/code/source/playfab/PlayFabEventPipeline.cpp
```diff
@@ -234,8 +234,8 @@ namespace PlayFab
         batchCounter++;
 
         this->batch.clear(); // batch vector will be reused
-        this->batch.reserve(this->settings->maximalNumberOfItemsInBatch);
-        if(this->settings->emitType == PlayFabEventPipelineType::PlayFabPlayStream)
+        this->batch.reserve(this->m_settings->maximalNumberOfItemsInBatch);
```
debhaldarMS: Please add a line before your conditional block begins - it makes the code easier to read

source/code/include/playfab/PlayFabAuthenticationContext.h
pgilmorepf: I did it to avoid warnings about member variables and function variables having the same name.
(I have my warning level set pretty high)
Do we do it elsewhere in XPlat - maybe, but probably not.
Do we do it elsewhere in other SDKs - yes.

---
In reply to: [343282558](https://github.com/PlayFab/XPlatCoreTemplate/pull/39#discussion_r343282558) [](ancestors = 343282558)

pgilmorepf:
> namespace PlayFab

Copied: Very much no. We might go so far as to completely disable that option...
Moved: Probably not.
Shared: Yes, we will very definitely need to have a single shared obj, referenced by multiple instanceAPIs.

---
In reply to: [550464267](https://github.com/PlayFab/XPlatCoreTemplate/pull/39#issuecomment-550464267) [](ancestors = 550464267)

---
Refers to: source/code/include/playfab/PlayFabAuthenticationContext.h:5 in 5fd376d. [](commit_id = 5fd376dd148fe94f311c25bcd19d895397ae3a30, deletion_comment = False)

source/code/include/playfab/PlayFabSettings.h
pgilmorepf: It's not unique, because it will definitely be simultaneously multi-referenced by simultaneous API calls.
Its contents will be modified by customers, but presumably the pointer/obj itself should never change.
Perhaps const static shared_ptr?  The PlayFabApiSettings inside must be mutable though.

---
In reply to: [343312976](https://github.com/PlayFab/XPlatCoreTemplate/pull/39#discussion_r343312976) [](ancestors = 343312976)

source/code/include/playfab/PlayFabError.h.ejs
```diff
@@ -42,4 +42,29 @@ namespace PlayFab
...
+    {
+    public:
+        PlayFabException() = delete;
+        PlayFabException(const PlayFabException& source) = default;
+        PlayFabException(PlayFabException&&) = default;
+        PlayFabException& operator=(const PlayFabException& source) = default;
+        PlayFabException& operator=(PlayFabException&& other) = default;
+
```
debhaldarMS: for my own knowledge - why do we need these i.e - they're not marked as deleted ?

debhaldarMS:
> namespace PlayFab

then please follow the rule of 5 for these

---
In reply to: [550491039](https://github.com/PlayFab/XPlatCoreTemplate/pull/39#issuecomment-550491039) [](ancestors = 550491039,550464267)

---
Refers to: source/code/include/playfab/PlayFabAuthenticationContext.h:5 in 5fd376d. [](commit_id = 5fd376dd148fe94f311c25bcd19d895397ae3a30, deletion_comment = False)

source/code/include/playfab/PlayFabError.h.ejs
pgilmorepf: std::runtime_error inherits from std::exception.
[http://www.cplusplus.com/reference/stdexcept/runtime_error/](http://www.cplusplus.com/reference/stdexcept/runtime_error/)
runtime_error is more appropriate for our exceptions, and in fact, the existing ones were already std::runtime_error


---
In reply to: [343288358](https://github.com/PlayFab/XPlatCoreTemplate/pull/39#discussion_r343288358) [](ancestors = 343288358)

source/code/include/playfab/PlayFabCallRequestContainerBase.h
pgilmorepf: I had a perfect storm of simultaneous name-collisons, because all of: local variables, function input params, and this class member var; to the point where this was the easiest and most clear resolution.

---
In reply to: [343287220](https://github.com/PlayFab/XPlatCoreTemplate/pull/39#discussion_r343287220) [](ancestors = 343287220)

source/code/include/playfab/PlayFabCallRequestContainerBase.h
pgilmorepf: RIGHT! This was an added method!
Oh, yes, I see though... it makes this class abstract...
This one is hard. Let's discuss in person.

---
In reply to: [343286948](https://github.com/PlayFab/XPlatCoreTemplate/pull/39#discussion_r343286948) [](ancestors = 343286948)

source/test/TestApp/PlayFabTestMultiUserInstance.cpp
pgilmorepf: My first resolution tried to use UNREFERENCED_PARAMETER.
UNREFERENCED_PARAMETER is not defined on all platforms, and was a compile error when I applied it trivially here.
I think I took one more step to attempt to also define it, and doing so also didn't trivially work, so I opted for commenting the input name as a solution we're already using in this codebase.

As for leaving the name unspecified: yeah, which is what this PR technically does by commenting it.
However, because it's a void*, it can be hard to figure out what it was supposed to be originally, if you remove the name.
It makes it harder to bring it back in a likely future where it's needed again.

---
In reply to: [343273143](https://github.com/PlayFab/XPlatCoreTemplate/pull/39#discussion_r343273143) [](ancestors = 343273143)

templates/PlayFab_InstanceApi.cpp.ejs
pgilmorepf: Take ownership: No, definitely not. context can be shared by any number of other instanceAPIs.
Because of life-cycle concerns, I'm hesitant to strip the obj out of the shared_ptr. This method could literally be the last ref to the given shared_ptr.
While I know that situation isn't well tested, and I think it'll fail for other reasons, I know for sure it's possible for this here to be the last shared_ptr ref to this obj, and the caller will lose their ref immediately after calling this.
In other words, I think it's worth investigating the case and testing/fixing where this is the last ref (as I'm pretty certain it'll crash), but under that circumstance, I'm pretty certain that this will need to be passed like this, in order to preserve the context obj.

---
In reply to: [343274785](https://github.com/PlayFab/XPlatCoreTemplate/pull/39#discussion_r343274785) [](ancestors = 343274785)

source/code/include/playfab/PlayFabApiSettings.h
pgilmorepf: Copied: Very much no. We might go so far as to completely disable that option...
Moved: Probably not.
Shared: Yes, we will very definitely need to have a single shared obj, referenced by multiple instanceAPIs.

---
In reply to: [343284614](https://github.com/PlayFab/XPlatCoreTemplate/pull/39#discussion_r343284614) [](ancestors = 343284614)

source/code/include/playfab/PlayFabSettings.h
pgilmorepf: 1. "Never worked" is true from the customer perspective.
2. The design of the underlying service feature changed literally the same week that we implemented this.
3. Agreed.

The history here is that it worked for a little while, while were were developing/testing it. Then we landed it. Then, the ability to access verticals directly was removed from the service level, right after we committed this. A right-hand, left-hand miscommunication between teams thing. But we'd already committed it, and it wasn't caught as broken until a week later. The automated testing DID catch it, but the "fix" was to disable the testing, because the feature was removed from the service.

---
In reply to: [343277480](https://github.com/PlayFab/XPlatCoreTemplate/pull/39#discussion_r343277480) [](ancestors = 343277480)

templates/PlayFab_InstanceApi.cpp.ejs
ScottMunroMS: Your statement that passing a raw_ptr would drop the reference count is not true (as long as you aren't misusing shared_ptr)

examples:
```
std::shared_ptr<t> sp = ...; // ref count 1
SomeFuncThatTakesSharedPointerByValue(sp); // ref count 2 while in the function, 1 upon returning assuming no one takes additional references inside of the function
```

```
std::shared_ptr<t> sp = ; // ref count 1
SomeFuncThatTakesRawPtr(sp.get()); // ref count stays 1 and the function has no opportunity to accidentally increase the reference count
```

There's a blog by Herb Sutter (one of the c++ big wigs) where he details the when it's appropriate to pass smart pointers around. the TL;DR is unless the callee is manipulating the ownership associated with the resource, pass by raw ptr or ref [https://herbsutter.com/2013/06/05/gotw-91-solution-smart-pointer-parameters/](https://herbsutter.com/2013/06/05/gotw-91-solution-smart-pointer-parameters/)

---
In reply to: [343327326](https://github.com/PlayFab/XPlatCoreTemplate/pull/39#discussion_r343327326) [](ancestors = 343327326,343274785)

source/code/source/playfab/PlayFabEventPipeline.cpp
pgilmorepf: Due to other comments, I've reverted all changes in this file.
So, I'm going to won't-fix this one, while acknowledging that you're totally right, but it's not right for this PR.

source/code/include/playfab/PlayFabError.h.ejs
pgilmorepf: The way we're throwing these, the copy constructor seems to be required. It's a compile error otherwise.
After that, nothing I'm doing is altering the way that runtime_error works, so I just want to let it keep working, with the one additional bit of info we're providing (PlayFabExceptionCode).

---
In reply to: [343317602](https://github.com/PlayFab/XPlatCoreTemplate/pull/39#discussion_r343317602) [](ancestors = 343317602)

source/code/include/playfab/PlayFabAuthenticationContext.h
debhaldarMS: We should figure out how to adhere to the same convention

---
In reply to: [343315277](https://github.com/PlayFab/XPlatCoreTemplate/pull/39#discussion_r343315277) [](ancestors = 343315277,343282558)

source/code/include/playfab/PlayFabError.h.ejs
debhaldarMS: You should provide explicit deletion of any of the copy control operators that you don't need

---
In reply to: [343340226](https://github.com/PlayFab/XPlatCoreTemplate/pull/39#discussion_r343340226) [](ancestors = 343340226,343317602)

source/code/include/playfab/PlayFabCallRequestContainerBase.h
pgilmorepf: I'm going to investigate whether this method and all its requirements should just be in the base class.
That'll require moving some things from the child to the base here, if I decide it's relevant.
If I go that route, this change will basically still be added, but it won't be abstract.

---
In reply to: [343322808](https://github.com/PlayFab/XPlatCoreTemplate/pull/39#discussion_r343322808) [](ancestors = 343322808,343286948)

source/code/include/playfab/PlayFabCallRequestContainerBase.h
pgilmorepf: As I said above, "apiSettings" and "settings" both conflict with the local, and funcion param vars, so I'm inclined to keep some kind of name change here.
They're protected fields, so it doesn't really affect the public interface... (GetApiSettings and GetContext being public).
So, despite the weirdness of m_ on two variables, I'm inclined to keep it, unless you ask me to revert it...

---
In reply to: [343322062](https://github.com/PlayFab/XPlatCoreTemplate/pull/39#discussion_r343322062) [](ancestors = 343322062,343287220)

PushpadantK: >m_settings [](start = 44, length = 10)

Is it then possible to make all member variables to be m_* for consistency?

source/code/include/playfab/PlayFabSettings.h
pgilmorepf: Yeah, const pointer to mutable object works, so that's been updated.

---
In reply to: [343317288](https://github.com/PlayFab/XPlatCoreTemplate/pull/39#discussion_r343317288) [](ancestors = 343317288,343312976)

source/code/include/playfab/PlayFabCallRequestContainer.h
```diff
@@ -15,11 +15,13 @@ namespace PlayFab
...
+            std::shared_ptr<PlayFabAuthenticationContext> context,
+            void* customData);
 
         virtual ~CallRequestContainer() override;
         std::string GetFullUrl() const;
+        virtual bool ValidateSettings() override;
```
PushpadantK: >virtual [](start = 8, length = 7)

I don't think virtual is needed here

source/code/source/playfab/PlayFabApiSettings.cpp
```diff
@@ -5,39 +5,33 @@
...
-        advertisingIdValue(PlayFabSettings::advertisingIdValue),
-        disableAdvertising(PlayFabSettings::disableAdvertising),
-#endif
-        verticalName(PlayFabSettings::verticalName),
         baseServiceHost(PlayFabSettings::productionEnvironmentURL),
-        titleId(PlayFabSettings::titleId)
+        titleId()
     {
+        // Don't let PlayFabSettings::staticSettings pull titleId from itself
+        if (PlayFabSettings::staticSettings != nullptr)
```
ScottMunroMS: in your latest commit, the shared_ptr is const, static, and initialized as non-null so I think this if-check is unnecessary. It should always be non-null

pgilmorepf: This is not QUITE true, because this constructor is called when that object is still being constructed.
That's why I had to add the check to begin with.

---
In reply to: [343379329](https://github.com/PlayFab/XPlatCoreTemplate/pull/39#discussion_r343379329) [](ancestors = 343379329)

source/code/include/playfab/PlayFabCallRequestContainer.h
pgilmorepf: You and Deb are absolutely right, this should not have been here. I was solving the immediate problem, but I forgot the purpose of the separation of BaseCRC (this one), and CRC, which inherits this one.
It's fixed now.

---
In reply to: [343375665](https://github.com/PlayFab/XPlatCoreTemplate/pull/39#discussion_r343375665) [](ancestors = 343375665)

source/code/include/playfab/PlayFabCallRequestContainerBase.h
pgilmorepf: Deb, you're right, ValidateSettings, and the settings, and the context should never have been in this file.
It's all removed.
Having PlayFab specifics of any kind in this file is a violation of the intent in the first place.
It's fully resolved now.

---
In reply to: [343371930](https://github.com/PlayFab/XPlatCoreTemplate/pull/39#discussion_r343371930) [](ancestors = 343371930,343322808,343286948)

source/code/include/playfab/PlayFabCallRequestContainerBase.h
pgilmorepf: That might be reasonable, but it'll also blow up this PR to be "almost every line in these files".
Separate PR?

---
In reply to: [343374534](https://github.com/PlayFab/XPlatCoreTemplate/pull/39#discussion_r343374534) [](ancestors = 343374534)

templates/PlayFab_Api.cpp.ejs
```diff
@@ -1,6 +1,7 @@
 #include <stdafx.h>
 
 <%- getApiDefine(api) %>
+#ifndef PLAYFAB_DISABLE_STATIC_API
```
PushpadantK: >PLAYFAB_DISABLE_STATIC_API [](start = 8, length = 26)

This is not defined at any project level or header so I assume we are not testing when this is defined? do we normally test preproccessor defines?

pgilmorepf: It was an unrelated idea-spark I had.
Since we don't define it, it is meaningless for now... but it could be a real thing soon enough...

---
In reply to: [343390440](https://github.com/PlayFab/XPlatCoreTemplate/pull/39#discussion_r343390440) [](ancestors = 343390440)

ScottMunroMS: unrelated idea-sparks that don't do anything and aren't tested sounds like something you don't want to check into the SDK. It smells like the beginning of something similar to the `verticalName` issue. It's code that's checked in, untested, and unused.

pgilmorepf: Quite right, removing now.

---
In reply to: [343391841](https://github.com/PlayFab/XPlatCoreTemplate/pull/39#discussion_r343391841) [](ancestors = 343391841)

source/code/include/playfab/PlayFabCallRequestContainerBase.h
debhaldarMS: that is going to break API

---
In reply to: [343389603](https://github.com/PlayFab/XPlatCoreTemplate/pull/39#discussion_r343389603) [](ancestors = 343389603,343374534)

templates/PlayFab_InstanceApi.cpp.ejs
pgilmorepf: We continued this conversation extensively in teams, and Todd & I discussed at length in person.
This is a really hard problem, since we abused shared_pointers to behave pretty close to the C# garbage collector.
It probably shouldn't have been designed this way, but it was, and we copied the C# design directly, rather than having a re-think for the C++ implementation.
This PR doesn't even really change the overall design/usage of shared_ptr regarding contexts & settings, except that it does it "a little more of the same".
So, while I "wontFix" it in this PR, it's not to say it isn't a problem.
At this point a re-think is going to be a customer breaking change, so we should keep that in mind, and schedule it accordingly.
Either way, I'd say it's a different task from this PR, which is just to make the current design work, without data-corruption and leaks.

---
In reply to: [343334818](https://github.com/PlayFab/XPlatCoreTemplate/pull/39#discussion_r343334818) [](ancestors = 343334818,343327326,343274785)

source/code/source/playfab/PlayFabIOSHttpPlugin.mm
```diff
@@ -91,38 +91,42 @@ void Cancel()
...
             }
-        } // UNLOCK httpRequestMutex
+            catch (const std::exception& ex)
+            {
+                PlayFabPluginManager::GetInstance().HandleException(ex);
+            }
+            catch (...)
+            {
+
+            }
```
debhaldarMS: add line before conditional statements

debhaldarMS:
>     struct PlayFabIOSHttpPlugin::RequestImpl

Have you validated this does not break ios C++ sdk ?

---
Refers to: source/code/source/playfab/PlayFabIOSHttpPlugin.mm:11 in b7c64dd. [](commit_id = b7c64ddb2653870e3b80968756ac8c8506b9ca95, deletion_comment = False)

source/code/source/playfab/PlayFabAndroidHttpPlugin.cpp
```diff
@@ -247,35 +247,39 @@ namespace PlayFab
 
     void PlayFabAndroidHttpPlugin::MakePostRequest(std::unique_ptr<CallRequestContainerBase> requestContainer)
```
debhaldarMS: Have you validated this does not break android C++ sdk ?

source/code/source/playfab/PlayFabAuthenticationContext.cpp
```diff
@@ -7,27 +7,42 @@ namespace PlayFab
...
-#if defined(ENABLE_PLAYFABSERVER_API) || defined(ENABLE_PLAYFABADMIN_API)
-        developerSecretKey.clear();
-#endif
     }
-}
\ No newline at end of file
+
+    void SetIfNotNull(const std::string& input, std::string& output)
+    {
+        if (!input.empty())
```
debhaldarMS: >if (!input.empty()) [](start = 8, length = 19)

what if input is null ? should we not check for it ?

ScottMunroMS: input is a std::string reference so there is no concern of null. A default constructed std::string will just be empty not null

pgilmorepf: Passed by ref, it shouldn't be null.
By the signature and usage here in this file, it should never be null.
It's technically possible, but you'd have to violate the signature of the function to accomplish it.

---
In reply to: [343405485](https://github.com/PlayFab/XPlatCoreTemplate/pull/39#discussion_r343405485) [](ancestors = 343405485)

source/code/source/playfab/PlayFabAndroidHttpPlugin.cpp
pgilmorepf: I have. I kicked off a run of the XPlat on Jenkins a couple times, whenever I changed the non-windows files, and (Win/iOS/Android) have passed.
The consoles won't be affected until we apply these changes to those templates.

---
In reply to: [343404709](https://github.com/PlayFab/XPlatCoreTemplate/pull/39#discussion_r343404709) [](ancestors = 343404709)

pgilmorepf:
>     struct PlayFabIOSHttpPlugin::RequestImpl

As discussed in the Android file: yes.

---
In reply to: [550570079](https://github.com/PlayFab/XPlatCoreTemplate/pull/39#issuecomment-550570079) [](ancestors = 550570079)

---
Refers to: source/code/source/playfab/PlayFabIOSHttpPlugin.mm:11 in b7c64dd. [](commit_id = b7c64ddb2653870e3b80968756ac8c8506b9ca95, deletion_comment = False)

source/code/source/playfab/PlayFabIOSHttpPlugin.mm
pgilmorepf: Generally, I prefer not to change lines that aren't part of this PR.
I'll do this one since it's adjacent to relevant changes and fast, but I'd resist doing lots of them.

---
In reply to: [343404249](https://github.com/PlayFab/XPlatCoreTemplate/pull/39#discussion_r343404249) [](ancestors = 343404249)

source/code/source/playfab/PlayFabIOSHttpPlugin.mm
```diff
@@ -91,38 +91,42 @@ void Cancel()
...
+            if(requestTask != nullptr)
+            { // LOCK httpRequestMutex
+                std::unique_lock<std::mutex> lock(httpRequestMutex);
+                requestTask->state = RequestTask::State::Pending;
+                pendingRequests.push_back(std::move(requestTask));
```
debhaldarMS: add a line break

source/code/source/playfab/PlayFabSettings.cpp.ejs
```diff
@@ -4,102 +4,24 @@
...
-            }
-            fullUrl += paramPair.first;
-            fullUrl += "=";
-            fullUrl += paramPair.second;
-        }
-        
-        return fullUrl;
-    }
-
-    bool PlayFabSettings::ValidateSettings(const std::shared_ptr<PlayFabAuthenticationContext> authenticationContext, const std::shared_ptr<PlayFabApiSettings> apiSettings, CallRequestContainer& container)
```
debhaldarMS: onsider renaming this because you're really modifying the state of CallRequestContainer object and not acting of APISettings.

PR #41, opened 2019-11-06
Removing excess std::moves

toddbellMSFT:
there should never be a need to std::move a shared_ptr. If there is a move requirement, possibly a shared_ptr is not the correct input.

This particular PR needs to hold off until Paul's changes are accepted here: https://github.com/PlayFab/XPlatCoreTemplate/pull/39 since there are a few places that already caught unneeded std::move calls.


source/code/include/playfab/PlayFabEventBuffer.h
```diff
@@ -20,7 +20,7 @@ namespace PlayFab
             next(nullptr),
             eventIndex(index),
             timestamp(std::time(nullptr)), // current time
-            eventRequest(std::move(request))
```
ScottMunroMS: >std::move( [](start = 25, length = 10)

the std::move actually isn't redundant here. std::shared_ptr can be copied so without std::move here we end up doing a shared_ptr copies which unnecessarily manipulates the ref count.

the std::move would only be unnecessary if you made the parameter to the constructor an r-value reference but I think what you had before is probably simpler

source/code/source/playfab/PlayFabEventBuffer.cpp
```diff
@@ -98,7 +98,7 @@ namespace PlayFab
         // create an event packet, set it to the tail->next and move the tail
 
         uint64_t currentEventIndex = eventIndex->fetch_add(1, std::memory_order_relaxed);
-        PlayFabEventPacket* event = CreateEventPacket(reinterpret_cast<uint8_t*>(eventStart), currentEventIndex, std::move(request));
+        PlayFabEventPacket* event = CreateEventPacket(reinterpret_cast<uint8_t*>(eventStart), currentEventIndex, request);
```
ScottMunroMS: >request [](start = 113, length = 7)

this std::move is not excessive and properly transfers ownership to CreateEventPacket. This does not need to be removed

source/code/include/playfab/PlayFabEventBuffer.h
toddbellMSFT: Since this is a shared_ptr, this std::move will ensure the original passed in request from the user would no longer be valid. Which brings up a question: Should we ever delete a user's shared_ptr? My only intention with shared_ptr is if we would share ownership, which kind of makes sense from the perspective of this object. (event request is also a shared_ptr, reinforcing this). 

BUT, this kind of ties into the const & idea. Herb Sutter did say passing by shared_ptr should only be by value if you care to transfer its ownership (This is what I meant earlier as making sense from this Object's perspective since we do want that ref count incremented by having this insternally as a shared_ptr). But const& if we don't really know if we will retain ownership, otherwise we should be passing the inner object's reference. 

So I think it is a real question as to why we should be passing this by a shared_ptr in the first place if ideally there's only ever one reference to that object (therefore this Packet should have contained a unique_ptr, which would make that std::move a more valid kind of thing). 

Sorry for the length, but this doesn't exactly end here though. This object is utilized by the PlayFabEventBuffer/Pipeline objects, which also prefer shared_ptr (which I think may increase the ref count around). 

Thoughts?

---
In reply to: [343929670](https://github.com/PlayFab/XPlatCoreTemplate/pull/41#discussion_r343929670) [](ancestors = 343929670)

ScottMunroMS: quick recap of offline chat we had:
* the rules for when to std::move vs not are a little soft. My personal preference is that the use of std::move or the absence of it for shared_ptrs should always indicate intent. std::move indicates "I am transferring ownership and no longer need to use this". a copy indicates "I am sharing ownership and still need to use this". In this way I think the std::move here should be retained. Herb Sutter's article did make a point about mentioning std::move-ing until you get to your destination. I think this operates under the assumption that the parameters that you are passing your shared_ptr through to its destination are r-value references (e.g. `std::shared_ptr<t>&&`). In which case that makes enough sense and the use of r-value references helps clarify when a callee is only intended to pass along a shared_ptr reference and not hold ownership of it.

* there was some discussion of using unique_ptr internally instead of std::shared_ptr. I am all onboard with using unique_ptr instead of shared_ptr because it means that there is somewhere a clear owner of the resource which makes that resource's lifetime easier to understand and way more predictable. I would be careful about doing it in such a way that isn't API breaking if possible. Changing from shared_ptr to unique_ptr may be a non-trivial task which requires some restructuring though since it may mean that some system which previously owned a resource may only have a weak reference to it once unique_ptr becomes your memory style.

* there was some discussion about the use of raw ptrs vs references. I think we are both onboard with using raw values to signal that a function does not have or manipulate ownership of a resource.

sounds like there is going to be a little bit of work before I continue to look at this PR so I will wait for the next iteration. 

---
In reply to: [343932433](https://github.com/PlayFab/XPlatCoreTemplate/pull/41#discussion_r343932433) [](ancestors = 343932433,343929670)

MichaelMackendMS:
>         DeleteEventPacket(headPtr);

is this excessive? it seems like it would avoids a refcount adjustment

---
Refers to: source/code/source/playfab/PlayFabEventBuffer.cpp:131 in 2ceb9f6. [](commit_id = 2ceb9f6c20f138f0c834c38483e882bba9084348, deletion_comment = False)

MichaelMackendMS:
>         PlayFabEventPacket* headPtr = head.load(std::memory_order_acquire);

I'm a little confused by the intent of all these atomics.  The fact that they're atomic would suggest that they might be manipulated concurrently.  Although the types are atomic, this function is not thread safe.  Is it expected to be thread safe?

---
Refers to: source/code/source/playfab/PlayFabEventBuffer.cpp:115 in 2ceb9f6. [](commit_id = 2ceb9f6c20f138f0c834c38483e882bba9084348, deletion_comment = False)

source/code/source/playfab/PlayFabEventPipeline.cpp
```diff
@@ -106,12 +106,19 @@ namespace PlayFab
...
             auto playFabEmitEventResponse = std::make_shared<PlayFabEmitEventResponse>();
             playFabEmitEventResponse->emitEventResult = emitResult;
 
+            // TODO: since we are std::move-ing the request,
+            //  we actually lose this info to return to the user.
+            // Do users care about this object?
+            // the entire point of this callback would be for users to retry if something went wrong,
+            // but with this particular change they must re-construct the request
+            // ( BUT which request? we no longer know...)
+            std::shared_ptr<PlayFabEmitEventRequest> req;
```
ScottMunroMS: >req [](start = 53, length = 3)

you are passing a null shared_ptr in the callback below... is that intentional? why aren't you passing the request object provided in the parameter? Should the code be updated so that ownership is only transferred when to the eventbuffer on success? otherwise I think we still want to own that object to pass back out in the CallbackRequest

source/code/source/playfab/PlayFabEventPipeline.cpp
```diff
@@ -106,12 +106,19 @@ namespace PlayFab
...
-            std::shared_ptr<const PlayFabEmitEventRequest> playFabEmitRequest = std::dynamic_pointer_cast<const PlayFabEmitEventRequest>(request);
             auto playFabEmitEventResponse = std::make_shared<PlayFabEmitEventResponse>();
             playFabEmitEventResponse->emitEventResult = emitResult;
 
+            // TODO: since we are std::move-ing the request,
+            //  we actually lose this info to return to the user.
+            // Do users care about this object?
+            // the entire point of this callback would be for users to retry if something went wrong,
+            // but with this particular change they must re-construct the request
+            // ( BUT which request? we no longer know...)
```
ScottMunroMS: this seems like a substantial open question and I have concerns about checking in changes while large open questions are still outstanding. What is the benefit of this change that is worth checking in with open questions like this.

ScottMunroMS:
I think it'd be great if you description was expanded upon a little bit.
What is the impact of this change? Why is it worth checking in? Will this have a positive impact on end-users? Is it just an internal refactor to keep developers righting better code? if so, how?

Originally this change was much more simple (just correcting usage of std::shared_ptr), but now it's a change in how lifetimes are managed inside the SDK and that's a more sizable change with more sizable risk to justify

source/test/TestApp/PlayFabEventTest.cpp
```diff
@@ -218,6 +218,16 @@ namespace PlayFabUnit
         std::bind(&PlayFabEventTest::NonStaticEmitEventCallback, this, std::placeholders::_1, std::placeholders::_2));
     }
 
+    void PlayFabEventTest::MutliThreadTest(TestContext& testContext)
+    {
+        //TODO: spin up X threads (2 or 10)
+        // Each thread should send 10 events
```
ScottMunroMS: considering the nature of this change, this TODO seems pretty critical to resolve before checking anything in.

ScottMunroMS:
>     }

I notice you don't have any tests that exercise all of this code with threadedCallbacks disabled. Early on during our SDK usage, the bumblelion team hit a crash from disabling threaded callbacks. Considering part of your change is to add a test for exercising lifetimes across different threading scenarios, adding a test that exercises `threadedCallbacks = false` seems really valuable

---
Refers to: source/test/TestApp/PlayFabEventTest.cpp:294 in c1466db. [](commit_id = c1466db639bc494d5f2178cb6778fb86618fd046, deletion_comment = False)

source/code/source/playfab/PlayFabEventApi.cpp
```diff
@@ -18,22 +18,22 @@ namespace PlayFab
 
     void PlayFabEventAPI::EmitEvent(std::unique_ptr<const IPlayFabEvent> event, const PlayFabEmitEventCallback callback) const
     {
-        auto eventRequest = std::shared_ptr<PlayFabEmitEventRequest>(new PlayFabEmitEventRequest());
+        auto eventRequest = std::unique_ptr<PlayFabEmitEventRequest>(new PlayFabEmitEventRequest());
```
ScottMunroMS: >std::unique_ptr<PlayFabEmitEventRequest>(new PlayFabEmitEventRequest()); [](start = 28, length = 72)

recommend std::make_unique everywhere you are using raw `new` to create a unique_ptr

e.g. `auto eventRequest = std::make_unique<PlayFabEmitEventRequest>();` it's shorter and prevents mismatching the types in the unique_ptr ctor and the `new` call

source/code/source/playfab/PlayFabEventPipeline.cpp
```diff
@@ -230,7 +237,7 @@ namespace PlayFab
 
         // add batch to flight tracking map
         void* customData = reinterpret_cast<void*>(batchCounter); // used to track batches across asynchronous Events API
-        this->batchesInFlight[customData] = std::move(this->batch);
+        this->batchesInFlight[customData] = this->batch;
```
ScottMunroMS: >this->batch; [](start = 44, length = 12)

this now copies the vector `this->batch`

is that intentional? it doesn't seem valuable since the batch is cleared and on line 243

I think what you want to do is keep the std::move here and then change the clear line on 243 to just assert that the size is 0

source/code/source/playfab/PlayFabEventPipeline.cpp
```diff
@@ -267,7 +274,7 @@ namespace PlayFab
             }
             else
             {
-                auto requestBatchPtr = std::shared_ptr<const std::vector<std::shared_ptr<const IPlayFabEmitEventRequest>>>(new std::vector<std::shared_ptr<const IPlayFabEmitEventRequest>>(std::move(foundBatchIterator->second)));
+                auto requestBatchPtr = std::shared_ptr<const std::vector<std::shared_ptr<const IPlayFabEmitEventRequest>>>(new std::vector<std::shared_ptr<const IPlayFabEmitEventRequest>>(foundBatchIterator->second));
```
ScottMunroMS: >foundBatchIterator [](start = 188, length = 18)

this is another spot where you are doing a copy instead of a move. is that intentional? I would think we'd want to avoid an entire vector copy

ScottMunroMS: >std::shared_ptr<const std::vector<std::shared_ptr<const IPlayFabEmitEventRequest>>> [](start = 39, length = 83)

this might just be a team code culture difference, but from my perspective this line is VERY hard to parse.

I think a good first step might be to use `std::make_shared` to simplify the `new` portion of this expression 

`std::make_shared<const std::vector<std::shared_ptr<const IPlayFabEmitEventRequest>>>(std::move(foundBatchIterator->second))`

in the longer term, it's pretty crazy to me that such a long type declaration is necessary and signals a pretty complicated interface. We had talked about it on friday, but this seems like a prime place to ask why all of the std::shared_ptr stuff is necessary. For now anything that can be done to make this stuff easier to read is goodness in my opinion!

ScottMunroMS:
>                     auto playFabEmitEventResponse = std::shared_ptr<PlayFabEmitEventResponse>(new PlayFabEmitEventResponse());

while you're here, I might recommend doing a sweep for places where it would simplify your code to use `std::make_shared` right now its use is inconsistent and it's harder to read the lines of code where it isn't used.

---
Refers to: source/code/source/playfab/PlayFabEventPipeline.cpp:283 in c1466db. [](commit_id = c1466db639bc494d5f2178cb6778fb86618fd046, deletion_comment = False)

source/code/source/playfab/PlayFabEventPipeline.cpp
```diff
@@ -341,12 +348,12 @@ namespace PlayFab
 
         if (playFabEmitRequest->callback != nullptr)
         {
-            playFabEmitRequest->callback(playFabEmitRequest->event, std::move(response));
+            playFabEmitRequest->callback(playFabEmitRequest->event, response);
         }
 
         if (playFabEmitRequest->stdCallback != nullptr)
```
ScottMunroMS: >stdCallback [](start = 32, length = 11)

what is the expectation around the use of `playFabEmitRequest->callback` vs `playFabEmitRequest->stdCallback` is there ever a world where a title would want to set both? is that intentionally permitted by your API?

source/code/source/playfab/PlayFabPluginManager.cpp
```diff
@@ -17,12 +17,12 @@ namespace PlayFab
 
     void PlayFabPluginManager::SetPlugin(std::shared_ptr<IPlayFabPlugin> plugin, const PlayFabPluginContract contract, const std::string& instanceName)
     {
-        GetInstance().SetPluginInternal(std::move(plugin), contract, instanceName);
```
ScottMunroMS: >std::move( [](start = 40, length = 10)

wouldn't call these std::moves excessive (per our discussion on friday)

source/code/source/playfab/PlayFabPluginManager.cpp
```diff
@@ -61,12 +61,12 @@ namespace PlayFab
         const auto pluginEntry = plugins.find(key);
         if (pluginEntry == plugins.end())
         {
```
ScottMunroMS: food for thought, have you considered opportunistically simplifying this method?

the `if end then {insert} else {delete and insert}` is unnecessarily verbose

```
key = ...;
plugins[key] = std::move(plugin);
```

is functionally equivalent and easier to read IMO

source/code/source/playfab/PlayFabWinHttpPlugin.cpp
```diff
@@ -212,7 +212,7 @@ namespace PlayFab
                         if (!GetBinaryPayload(reqContainer, payload, payloadSize))
```
ScottMunroMS: >GetBinaryPayload [](start = 29, length = 16)

unrelated to your current change, but while reasoning about the flow of this code I inspected GetBinaryPayload and noticed that it was a no-op that always returns false making this code more complicated than it needs to be. If there's expected work to implement getbinarypayload I recommend adding a TODO to GetBinaryPayload. If not, I recommend making a work item to prune this unnecessary code

toddbellMSFT:
After talking with a few folks. The scope of the fix has changed.

PR #43, opened 2019-11-08
Move all platform-specific timestamp manipulation to a single new PlatformUtils.h file

pgilmorepf:
All other files cross reference this file, and it's generic interface
A few header cleanups and improvements that were close to this change - Particularly about being #defined against platforms
Converting several non-customer facing time calculations to use std::chrono

Repo changes applied by these template changes viewable here:
https://github.com/PlayFab/XPlatCppSdk/pull/32

### DIFF
--- a/source/code/include/playfab/PlayFabBaseModel.h
+++ b/source/code/include/playfab/PlayFabBaseModel.h
@@ -1,38 +1,18 @@
 #pragma once
 
-#include <ctime>
+#include <assert.h>
 #include <functional>
 #include <list>
 #include <map>
-#include <assert.h>
-
-#include <sstream>
-#include <iomanip>
-#include <memory>
 
 #include <playfab/PlayFabPlatformMacros.h>
 #include <playfab/PlayFabJsonHeaders.h>
-#include <playfab/PlayFabAuthenticationContext.h>
+#include <playfab/PlayFabPlatformUtils.h>
 
 namespace PlayFab
 {
-#if defined(PLAYFAB_PLATFORM_WINDOWS) || defined(PLAYFAB_PLATFORM_XBOX)
-    typedef signed __int64 Int64;
-    typedef signed __int32 Int32;
-    typedef signed __int16 Int16;
+    class PlayFabAuthenticationContext;
 
-    typedef unsigned __int64 Uint64;
-    typedef unsigned __int32 Uint32;
-    typedef unsigned __int16 Uint16;
-#elif defined(PLAYFAB_PLATFORM_LINUX) || defined(PLAYFAB_PLATFORM_IOS) || defined(PLAYFAB_PLATFORM_ANDROID) || defined(PLAYFAB_PLATFORM_PLAYSTATION)
-    typedef int64_t Int64;
-    typedef int32_t Int32;
-    typedef int16_t Int16;
-
-    typedef uint64_t Uint64;
-    typedef uint32_t Uint32;
-    typedef uint16_t Uint16;
-#endif
     template <typename BoxedType>
     class Boxed
     {
@@ -97,35 +77,12 @@ namespace PlayFab
     // Utilities for [de]serializing time_t to/from json
     inline void ToJsonUtilT(const time_t input, Json::Value& output)
     {
-        tm timeInfo;
-#if defined(PLAYFAB_PLATFORM_WINDOWS) || defined(PLAYFAB_PLATFORM_XBOX)
-        gmtime_s(&timeInfo, &input);
-#elif defined(PLAYFAB_PLATFORM_LINUX) || defined(PLAYFAB_PLATFORM_IOS) || defined(PLAYFAB_PLATFORM_ANDROID) || defined(PLAYFAB_PLATFORM_PLAYSTATION)
-        timeInfo = *gmtime(&input);
-#endif
-        char buff[40];
-        strftime(buff, 40, "%Y-%m-%dT%H:%M:%S.000Z", &timeInfo);
-        output = Json::Value(buff);
+        output = Json::Value(LocalTimeTToUtcString(input));
     }
-
     inline void FromJsonUtilT(const Json::Value& input, time_t& output)
     {
-        if (input == Json::Value::null)
-        {
-            return;
-        }
-
-        const std::string timeStr = input.asString();
-        tm timeStruct = {};
-        std::istringstream iss(timeStr);
-        iss >> std::get_time(&timeStruct, "%Y-%m-%dT%T");
-#if defined(PLAYFAB_PLATFORM_PLAYSTATION)
-        output = mktime(&timeStruct);
-#elif defined(PLAYFAB_PLATFORM_IOS) || defined(PLAYFAB_PLATFORM_ANDROID) || defined(PLAYFAB_PLATFORM_LINUX)
-        output = timegm(&timeStruct);
-#else
-        output = _mkgmtime(&timeStruct);
-#endif
+        if (input == Json::Value::null) return;
+        output = UtcStringToLocalTimeT(input.asString());
     }
 
     inline void ToJsonUtilT(const Boxed<time_t>& input, Json::Value& output)
--- a/source/code/include/playfab/PlayFabBaseModel.h
+++ b/source/code/include/playfab/PlayFabBaseModel.h
@@ -4,6 +4,7 @@
 #include <functional>
 #include <list>
 #include <map>
+#include <memory>
 
 #include <playfab/PlayFabPlatformMacros.h>
 #include <playfab/PlayFabJsonHeaders.h>
--- a/source/code/include/playfab/PlayFabPlatformMacros.h
+++ b/source/code/include/playfab/PlayFabPlatformMacros.h
@@ -40,6 +40,8 @@ typedef unsigned __int64 Uint64;
 typedef unsigned __int32 Uint32;
 typedef unsigned __int16 Uint16;
 #elif defined(PLAYFAB_PLATFORM_LINUX) || defined(PLAYFAB_PLATFORM_IOS) || defined(PLAYFAB_PLATFORM_ANDROID) || defined(PLAYFAB_PLATFORM_PLAYSTATION)
+#include <cstdint>
+
 typedef int64_t Int64;
 typedef int32_t Int32;
 typedef int16_t Int16;
--- a/source/code/include/playfab/PlayFabPlatformMacros.h
+++ b/source/code/include/playfab/PlayFabPlatformMacros.h
@@ -30,3 +30,21 @@
 #if defined(_WIN32) && !defined(_DURANGO)
 #define PLAYFAB_PLATFORM_WINDOWS
 #endif //_WIN32
+
+#if defined(PLAYFAB_PLATFORM_WINDOWS) || defined(PLAYFAB_PLATFORM_XBOX)
+typedef signed __int64 Int64;
+typedef signed __int32 Int32;
+typedef signed __int16 Int16;
+
+typedef unsigned __int64 Uint64;
+typedef unsigned __int32 Uint32;
+typedef unsigned __int16 Uint16;
+#elif defined(PLAYFAB_PLATFORM_LINUX) || defined(PLAYFAB_PLATFORM_IOS) || defined(PLAYFAB_PLATFORM_ANDROID) || defined(PLAYFAB_PLATFORM_PLAYSTATION)
+typedef int64_t Int64;
+typedef int32_t Int32;
+typedef int16_t Int16;
+
+typedef uint64_t Uint64;
+typedef uint32_t Uint32;
+typedef uint16_t Uint16;
+#endif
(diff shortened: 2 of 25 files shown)


source/code/include/playfab/PlayFabPlatformUtils.h
```diff
@@ -0,0 +1,78 @@
...
+#include <chrono>
+#include <ctime>
+#include <iomanip>
+#include <sstream>
+
+namespace PlayFab
+{
+    typedef std::chrono::time_point<std::chrono::system_clock> TimePoint;
+
+    inline TimePoint GetPlayFabTimePointNow()
```
ScottMunroMS: >PlayFabTime [](start = 24, length = 11)

nit:  I think the playfab prefix is unnecessary. recommend just GetTimePointNow()

source/code/include/playfab/PlayFabPlatformUtils.h
```diff
@@ -0,0 +1,78 @@
...
+    {
+        return std::chrono::system_clock::now();
+    }
+
+    inline time_t GetPlayFabTimeTNow()
+    {
+        time_t now = std::chrono::system_clock::to_time_t(GetPlayFabTimePointNow());
+#if defined(PLAYFAB_PLATFORM_IOS) || defined(PLAYFAB_PLATFORM_ANDROID) || defined(PLAYFAB_PLATFORM_LINUX) || defined(PLAYFAB_PLATFORM_SWITCH)
+        return static_cast<Json::Int64>(now);
```
ScottMunroMS: >static_cast<Json::Int64> [](start = 15, length = 24)

the return type is time_t which `now` is already typed as. what is this static_cast for?

source/code/include/playfab/PlayFabPlatformUtils.h
```diff
@@ -0,0 +1,78 @@
...
+#else // PLAYFAB_PLATFORM_IOS || PLAYFAB_PLATFORM_ANDROID || PLAYFAB_PLATFORM_LINUX || PLAYFAB_PLATFORM_SWITCH
+        return now;
+#endif // PLAYFAB_PLATFORM_IOS || PLAYFAB_PLATFORM_ANDROID || PLAYFAB_PLATFORM_LINUX || PLAYFAB_PLATFORM_SWITCH
+    }
+
+    inline void AppendIntToString(int value, std::string& output)
```
ScottMunroMS: >AppendIntToString [](start = 16, length = 17)

this method only appears to be used in the test app.

instead of adding an SDK platform helper, why not just use the native std::to_string?

source/code/include/playfab/PlayFabPlatformUtils.h
```diff
@@ -0,0 +1,78 @@
...
+        char buffer[16];
+#if defined(PLAYFAB_PLATFORM_IOS) || defined(PLAYFAB_PLATFORM_ANDROID) || defined(PLAYFAB_PLATFORM_LINUX) || defined(PLAYFAB_PLATFORM_PLAYSTATION) || defined(PLAYFAB_PLATFORM_SWITCH)
+        sprintf(buffer, "%d", value);
+#else // PLAYFAB_PLATFORM_IOS || PLAYFAB_PLATFORM_ANDROID || PLAYFAB_PLATFORM_LINUX || PLAYFAB_PLATFORM_PLAYSTATION || PLAYFAB_PLATFORM_SWITCH
+        sprintf_s(buffer, "%d", value);
+#endif // PLAYFAB_PLATFORM_IOS || PLAYFAB_PLATFORM_ANDROID || PLAYFAB_PLATFORM_LINUX || PLAYFAB_PLATFORM_PLAYSTATION || PLAYFAB_PLATFORM_SWITCH
+        output.append(buffer);
+    }
+
+    inline std::string LocalTimeTToUtcString(time_t now)
```
ScottMunroMS: >LocalTime [](start = 23, length = 9)

you use the term "LocalTime" to describe time_t in a number of places. time_t does not represent a local time. it's just an opaque "time" value that usually represents number of seconds since the unix epoch which would not be a locally defined value. Recommend dropping the "local" where appropriate and renaming accordingly. e.g. here I think a better name would just be `TimeAsString` or `TimeToUtcString` or something

source/code/include/playfab/PlayFabPlatformUtils.h
```diff
@@ -0,0 +1,78 @@
...
+    {
+        std::string output;
+
```
ScottMunroMS: this temporary is unnecessary. you could just return `buff` at the end

source/code/include/playfab/PlayFabPlatformUtils.h
```diff
@@ -0,0 +1,78 @@
...
+        tm timeInfo;
+#if defined(PLAYFAB_PLATFORM_WINDOWS) || defined(PLAYFAB_PLATFORM_XBOX)
+        gmtime_s(&timeInfo, &now);
```
ScottMunroMS: >gmtime_s [](start = 8, length = 8)

gmtime_s returns an error value of type errno_t. I recommend asserting that the error value indicates success

source/code/include/playfab/PlayFabPlatformUtils.h
```diff
@@ -0,0 +1,78 @@
...
+#elif defined(PLAYFAB_PLATFORM_LINUX) || defined(PLAYFAB_PLATFORM_IOS) || defined(PLAYFAB_PLATFORM_ANDROID) || defined(PLAYFAB_PLATFORM_PLAYSTATION)
+        timeInfo = *gmtime(&input);
```
ScottMunroMS: >gmtime [](start = 20, length = 6)

this function isn't reentrant or threadsafe I'm pretty sure. you should find a thread-safe/reentrant equivalent. I think that would be gmtime_r

source/code/include/playfab/PlayFabPlatformUtils.h
```diff
@@ -0,0 +1,78 @@
...
+        return LocalTimeTToUtcString(std::chrono::system_clock::to_time_t(now));
+    }
+
+    inline time_t UtcStringToLocalTimeT(std::string utcString)
+    {
+        time_t output;
+        tm timeStruct = {};
+
+        std::istringstream iss(utcString);
+        iss >> std::get_time(&timeStruct, "%Y-%m-%dT%T");
```
ScottMunroMS: >"%Y-%m-%dT%T" [](start = 42, length = 13)

shouldn't the format here and in LocalTimeToUtcString be identical?

source/code/include/playfab/PlayFabPlatformUtils.h
```diff
@@ -0,0 +1,78 @@
...
+
+        return output;
+    }
+
+    inline std::string LocalTimePointToUtcString(TimePoint now)
+    {
+        return LocalTimeTToUtcString(std::chrono::system_clock::to_time_t(now));
+    }
+
+    inline time_t UtcStringToLocalTimeT(std::string utcString)
```
ScottMunroMS: >std::string [](start = 40, length = 11)

const-ref to avoid copy

source/code/include/playfab/PlayFabPlatformMacros.h
```diff
@@ -29,4 +29,22 @@
...
+#endif //_WIN32
+
+#if defined(PLAYFAB_PLATFORM_WINDOWS) || defined(PLAYFAB_PLATFORM_XBOX)
+typedef signed __int64 Int64;
+typedef signed __int32 Int32;
+typedef signed __int16 Int16;
+
+typedef unsigned __int64 Uint64;
+typedef unsigned __int32 Uint32;
+typedef unsigned __int16 Uint16;
```
ScottMunroMS: why is this typedef necessary? windows and xbox have access to the stdint.h header so they can use types like uint64_t

this has bothered me before when looking at sdk code. it'd be great to get rid of these typedefs if possible

source/code/include/playfab/PlayFabBaseModel.h
```diff
@@ -97,35 +77,12 @@ namespace PlayFab
...
-        std::istringstream iss(timeStr);
-        iss >> std::get_time(&timeStruct, "%Y-%m-%dT%T");
-#if defined(PLAYFAB_PLATFORM_PLAYSTATION)
-        output = mktime(&timeStruct);
-#elif defined(PLAYFAB_PLATFORM_IOS) || defined(PLAYFAB_PLATFORM_ANDROID) || defined(PLAYFAB_PLATFORM_LINUX)
-        output = timegm(&timeStruct);
-#else
-        output = _mkgmtime(&timeStruct);
-#endif
+        if (input == Json::Value::null) return;
```
ScottMunroMS: > return; [](start = 39, length = 8)

nit: I think I've seen deb specifically request braces on all if-else statements even if they are one line. if that's true you should fix this (it's certainly my preferred style)

source/test/TestApp/PlayFabApiTest.cpp
```diff
@@ -35,12 +38,12 @@ namespace PlayFabUnit
 
     void PlayFabApiTest::OnErrorSharedCallback(const PlayFabError& error, void* customData)
     {
-        TestContext* testContext = reinterpret_cast<TestContext*>(customData);
+        TestContext* testContext = static_cast<TestContext*>(customData);
```
ScottMunroMS: >static_cast [](start = 35, length = 11)

nice!

source/code/source/playfab/PlayFabEventBuffer.cpp
```diff
@@ -27,8 +27,8 @@ namespace PlayFab
         bufferArray(std::unique_ptr<uint8_t[]>(new uint8_t[buffMask + 1])),
         buffStart((uint64_t)(bufferArray.get())),
         buffEnd(buffStart + buffMask + 1),
-        eventIndex(std::shared_ptr<std::atomic<uint64_t>>(new std::atomic<uint64_t>(0))),
-        disabled(false)
+        disabled(false),
+        eventIndex(std::shared_ptr<std::atomic<uint64_t>>(new std::atomic<uint64_t>(0)))
```
ScottMunroMS: >std::shared_ptr<std::atomic<uint64_t>>(new std::atomic<uint64_t>(0)) [](start = 19, length = 68)

nit: this is unnecessarily verbose.

recommend `std::make_shared<std::atomic_uint64_t>(0)`

source/code/include/playfab/PlayFabPlatformUtils.h
pgilmorepf: No idea. I haven't tested it on those platforms yet.
I'm just copy pasting the code that was already there.

source/code/include/playfab/PlayFabPlatformUtils.h
pgilmorepf: std::to_string doesn't exist on all platforms.

source/code/include/playfab/PlayFabPlatformUtils.h
pgilmorepf: While I generally agree, our basic issue is that we don't have any consistent plan for doing asserts in the SDK. I don't want to further complicate this already difficult PR with a whole new feature.

source/code/include/playfab/PlayFabPlatformUtils.h
pgilmorepf: That's worth noting, but changing the code isn't part of this PR. I'm trying to move the platform specific time code into PlatformUtils, so that suggestions above are easier to resolve.
I'm willing to file a bug on it though.

source/code/include/playfab/PlayFabPlatformUtils.h
pgilmorepf: Filed task.

source/code/include/playfab/PlayFabPlatformUtils.h
pgilmorepf: Filed bug.

source/code/include/playfab/PlayFabPlatformMacros.h
pgilmorepf: I didn't define it, I just moved it to the platform-specific file, because it clearly is.
I think the intent is to have a consistent label for the strict-types, and they aren't platform consistent.
I'm willing to improve this, but not in this PR.

source/code/include/playfab/PlayFabPlatformUtils.h
pgilmorepf: Bug filed.

source/code/include/playfab/PlayFabPlatformUtils.h
pgilmorepf: This is probably a design problem that we have to resolve.
In my past experiments, parsing and printing the output of time_t required shifting/correcting for the timezone, given the input string. If you didn't, the result time_t didn't represent "now" correctly.
There's a unit test that verifies that when we retrieve the server GetTime, it is correctly parsed into a time_t which matches the local clock.
I don't care much about the names of these methods, but I do want to make sure that our interface works.
Once we've verified the interface is working as expected on all platforms, I'll clean up the function names and look for unused code.

PR #44, opened 2019-11-11
Adding threading tests for the event pipeline

toddbellMSFT:
We are considering a refactor of the PlayFabEventPipeline which is expected to be threadsafe. We don't have tests that cover this scenario, so I added 3 tests: 1 basic 5 threads with 6 events per thread, and then a 30 thread with 2 events per and test with 2 threads and 15 events per.

### DIFF
--- a/source/test/TestApp/PlayFabEventTest.cpp
+++ b/source/test/TestApp/PlayFabEventTest.cpp
@@ -60,12 +60,12 @@ namespace PlayFabUnit
         EventsModels::WriteEventsRequest request;
 
         // send several events
-        for (int i = 0; i < 2; i++)
-        {
-            request.Events.push_back(CreateEventContents("event_A_", i));
-            request.Events.push_back(CreateEventContents("event_B_", i));
-        }
-
+        // for (int i = 0; i < 2; i++)
+        // {
+             request.Events.push_back(CreateEventContents("event_A_", 0));
+        //     request.Events.push_back(CreateEventContents("event_B_", i));
+        // }
+        
         // PlayFabEventsAPI::WriteEvents(request,
         //     &PlayFabEventTest::OnEventsApiSucceeded,
         //     &PlayFabEventTest::OnEventsApiFailed,
--- a/source/test/TestApp/PlayFabEventTest.cpp
+++ b/source/test/TestApp/PlayFabEventTest.cpp
@@ -210,11 +210,12 @@ namespace PlayFabUnit
 
     void PlayFabEventTest::PrivateMemberCallbackTest(TestContext& testContext)
     {
-        eventTestContext = std::make_shared<TestContext*>(&testContext);
+       eventTestContext = std::make_shared<TestContext*>(&testContext);
 
-        std::shared_ptr<PlayFabEventAPI*> api = SetupEventTest();
+        //std::shared_ptr<PlayFabEventAPI*> api = SetupEventTest();
+        eventApi = SetupEventTest();
 
-        (*api)->EmitEvent(MakeEvent(0, PlayFabEventType::Default),
+        (*eventApi)->EmitEvent(MakeEvent(0, PlayFabEventType::Default),
         std::bind(&PlayFabEventTest::NonStaticEmitEventCallback, this, std::placeholders::_1, std::placeholders::_2));
     }
 
--- a/source/test/TestApp/PlayFabEventTest.cpp
+++ b/source/test/TestApp/PlayFabEventTest.cpp
@@ -99,7 +99,7 @@ namespace PlayFabUnit
         (*eventTestContext)->Pass("Private member called back!");
     }
 
-    void PlayFabEventTest::MyThreadingCallback(std::shared_ptr<const PlayFab::IPlayFabEvent> event, std::shared_ptr<const PlayFab::IPlayFabEmitEventResponse> response)
+    void PlayFabEventTest::MyThreadingCallback(std::shared_ptr<const PlayFab::IPlayFabEvent> /*event*/, std::shared_ptr<const PlayFab::IPlayFabEmitEventResponse> /*response*/)
     {
         numEventsHeard++;
         if(numEventsHeard == numTotalThreadedEvents)
--- a/source/test/TestApp/PlayFabEventTest.cpp
+++ b/source/test/TestApp/PlayFabEventTest.cpp
@@ -281,7 +281,7 @@ namespace PlayFabUnit
         AddTest("HeavyweightEvents", &PlayFabEventTest::HeavyweightEvents);
         AddTest("LightweightEvents", &PlayFabEventTest::LightweightEvents);
         AddTest("LambdaCallback", &PlayFabEventTest::LambdaCallbackTest);
-        //AddTest("PrivateMemberCallback", &PlayFabEventTest::PrivateMemberCallbackTest);
+        AddTest("PrivateMemberCallback", &PlayFabEventTest::PrivateMemberCallbackTest);
         AddTest("BasicMultiThreadedTest", &PlayFabEventTest::BasicMultiThreadedTest);
         AddTest("ManyThreadsLowEventsPerTest", &PlayFabEventTest::ManyThreadsLowEventsPerTest);
         AddTest("FewThreadsHighEventsPerTest", &PlayFabEventTest::FewThreadsHighEventsPerTest);
--- a/source/test/TestApp/PlayFabEventTest.cpp
+++ b/source/test/TestApp/PlayFabEventTest.cpp
@@ -277,11 +277,11 @@ namespace PlayFabUnit
 #if !defined(PLAYFAB_PLATFORM_IOS) && !defined(PLAYFAB_PLATFORM_ANDROID) && !defined(PLAYFAB_PLATFORM_PLAYSTATION) && !defined(PLAYFAB_PLATFORM_SWITCH)
         AddTest("QosResultApi", &PlayFabEventTest::QosResultApi);
 #endif
-        AddTest("EventsApi", &PlayFabEventTest::EventsApi);
+        //AddTest("EventsApi", &PlayFabEventTest::EventsApi);
         AddTest("HeavyweightEvents", &PlayFabEventTest::HeavyweightEvents);
         AddTest("LightweightEvents", &PlayFabEventTest::LightweightEvents);
         AddTest("LambdaCallback", &PlayFabEventTest::LambdaCallbackTest);
-        AddTest("PrivateMemberCallback", &PlayFabEventTest::PrivateMemberCallbackTest);
+        //AddTest("PrivateMemberCallback", &PlayFabEventTest::PrivateMemberCallbackTest);
         AddTest("BasicMultiThreadedTest", &PlayFabEventTest::BasicMultiThreadedTest);
         AddTest("ManyThreadsLowEventsPerTest", &PlayFabEventTest::ManyThreadsLowEventsPerTest);
         AddTest("FewThreadsHighEventsPerTest", &PlayFabEventTest::FewThreadsHighEventsPerTest);
--- a/source/test/TestApp/PlayFabEventTest.cpp
+++ b/source/test/TestApp/PlayFabEventTest.cpp
@@ -59,24 +59,17 @@ namespace PlayFabUnit
 
         EventsModels::WriteEventsRequest request;
 
+        // TODO: Bug 38165 automated builds either time out here or tests will never complete.
         // send several events
-        //for (int i = 0; i < 1; i++)
+        //for (int i = 0; i < 2; i++)
         //{
             request.Events.push_back(CreateEventContents("event_A_", 0));
             //request.Events.push_back(CreateEventContents("event_B_", i));
         //}
-        
+
         PlayFabEventsAPI::WriteEvents(request,
-            [](const PlayFab::EventsModels::WriteEventsResponse, void* customData)
-            {
-                TestContext* testContext = reinterpret_cast<TestContext*>(customData);
-                testContext->Pass();
-            },
-            [](const PlayFab::PlayFabError& error, void* customData)
-            {
-                TestContext* testContext = reinterpret_cast<TestContext*>(customData);
-                testContext->Fail(error.GenerateErrorReport());
-            },
+            Callback(&PlayFabEventTest::OnEventsApiSucceeded),
+            Callback(&PlayFabEventTest::OnEventsApiFailed),
             &testContext);
     }
 
--- a/source/test/TestApp/PlayFabEventTest.cpp
+++ b/source/test/TestApp/PlayFabEventTest.cpp
@@ -234,8 +234,7 @@ namespace PlayFabUnit
                             [&eventsRemaining, pNumEventsPerThread]
                             (std::shared_ptr<const PlayFab::IPlayFabEvent>, std::shared_ptr<const PlayFab::IPlayFabEmitEventResponse>)
                             {
-                                eventsRemaining--;
-                                if (eventsRemaining == 0)
+                                if (--eventsRemaining == 0)
                                 {
                                     (*eventTestContext)->Pass("Threaded callback Received all events Emitted.");
                                 }
--- a/source/test/TestApp/PlayFabEventTest.cpp
+++ b/source/test/TestApp/PlayFabEventTest.cpp
@@ -60,16 +60,12 @@ namespace PlayFabUnit
         EventsModels::WriteEventsRequest request;
 
         // send several events
-        // for (int i = 0; i < 2; i++)
-        // {
-             request.Events.push_back(CreateEventContents("event_A_", 0));
-        //     request.Events.push_back(CreateEventContents("event_B_", i));
-        // }
+        for (int i = 0; i < 1; i++)
+        {
+            request.Events.push_back(CreateEventContents("event_A_", i));
+            request.Events.push_back(CreateEventContents("event_B_", i));
+        }
         
-        // PlayFabEventsAPI::WriteEvents(request,
-        //     &PlayFabEventTest::OnEventsApiSucceeded,
-        //     &PlayFabEventTest::OnEventsApiFailed,
-        //     &testContext);
         PlayFabEventsAPI::WriteEvents(request,
             [](const PlayFab::EventsModels::WriteEventsResponse, void* customData)
             {
--- a/source/test/TestApp/PlayFabEventTest.cpp
+++ b/source/test/TestApp/PlayFabEventTest.cpp
@@ -92,8 +92,6 @@ namespace PlayFabUnit
     int PlayFabEventTest::eventFailCount;
     std::string PlayFabEventTest::eventFailLog;
 
-    std::vector<std::thread> PlayFabEventTest::testThreadPool;
-
     void PlayFabEventTest::NonStaticEmitEventCallback(std::shared_ptr<const PlayFab::IPlayFabEvent> /*event*/, std::shared_ptr<const PlayFab::IPlayFabEmitEventResponse> /*response*/)
     {
         (*eventTestContext)->Pass("Private member called back!");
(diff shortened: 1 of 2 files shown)


toddbellMSFT:
These new tests pass on Jenkins for Windows: https://pf-jenkins.playfab.com/view/SDK/job/SDK_Work_XPlat/4378/console

Android and iOS tests: https://pf-jenkins.playfab.com/view/SDK/job/SDK_Work_XPlat/4379/console

source/test/TestApp/PlayFabEventTest.h
```diff
@@ -57,11 +64,19 @@ namespace PlayFabUnit
             static int eventFailCount;
             static std::string eventFailLog;
 
+            static const unsigned int numThreads = 6;
```
ScottMunroMS: >unsigned int [](start = 25, length = 12)

nit: instead of using `unsigned int`, it would be shorter (i.e. easier to read) to use a sized-int type ala uint32_t

source/test/TestApp/PlayFabEventTest.h
```diff
@@ -47,6 +47,13 @@ namespace PlayFabUnit
             void LambdaCallbackTest(TestContext& testContext);
             void PrivateMemberCallbackTest(TestContext& testContext);
 
+            void BasicMultiThreadedTest(TestContext& testContext);
+
+            // We need to make sure these are able to pass within our TIMEOUT limits.
```
ScottMunroMS: >TIMEOUT [](start = 70, length = 7)

nit: is there a reason this is all caps? does it correlate to some constant somewhere? seems a bit weird to me but maybe it's communicating something I'm not seeing

source/test/TestApp/PlayFabEventTest.cpp
```diff
@@ -218,6 +229,40 @@ namespace PlayFabUnit
         std::bind(&PlayFabEventTest::NonStaticEmitEventCallback, this, std::placeholders::_1, std::placeholders::_2));
     }
 
+    void PlayFabEventTest::GenericMultiThreadedTest(unsigned int pNumThreads, unsigned int pNumEventsPerThread)
+    {
+        numEventsHeard = 0;
+        for (unsigned int thread = 0; thread < pNumThreads; ++thread)
+        {
+            testThreadPool.emplace_back(std::thread([pNumEventsPerThread, this](){
```
ScottMunroMS: >emplace_back(std::thread [](start = 27, length = 24)

the whole point of emplace_back is to pass the args that are forwarded to the constructor so it doesn't make sense to call emplace_back with an already constructed value.

e.g.
```
std::vector<Type> x;
x.emplace_back( X(arg1, arg2, arg3) ); // wrong/unnecessary
x.emplace_back(arg1, arg2, arg3); // correct use of emplace_back
x.push_back(X(arg1, arg2, arg3)); // if you want the value passed into the vector to explicitly have the type construction somewhere, push_back would be the method to use
```

ScottMunroMS:
>     std::shared_ptr<PlayFabEventAPI*> PlayFabEventTest::SetupEventTest(int maxBatchWaitTime, int maxItemsInBatch, int maxBatchesInFlight)

why does this return a shared_ptr to a pointer?

I don't see any raw deletes in here so I think your tests leak memory and make the code more complicated than it needs to be.

if this was just std::shared_ptr<PlayFabEventAPI> then once all references to this go away the API object would clean up. It would also make your tests easier to read as every LoC using `api` could be shorted from `(*api)->` to just `api->`


---
Refers to: source/test/TestApp/PlayFabEventTest.cpp:355 in 28a9118. [](commit_id = 28a911864c0819fb47cc78e81aca866a6ece7688, deletion_comment = False)

ScottMunroMS:
>     std::shared_ptr<PlayFabEventAPI*> PlayFabEventTest::SetupEventTest(int maxBatchWaitTime, int maxItemsInBatch, int maxBatchesInFlight)

actually this is likely even more concerning than I had originally thought because the fact that this object is being leaked means you are never testing the PlayFabEventAPI destructor behavior

---
In reply to: [553106455](https://github.com/PlayFab/XPlatCoreTemplate/pull/44#issuecomment-553106455) [](ancestors = 553106455)

---
Refers to: source/test/TestApp/PlayFabEventTest.cpp:355 in 28a9118. [](commit_id = 28a911864c0819fb47cc78e81aca866a6ece7688, deletion_comment = False)

ScottMunroMS:
>     std::shared_ptr<PlayFabEventAPI*> PlayFabEventTest::SetupEventTest(int maxBatchWaitTime, int maxItemsInBatch, int maxBatchesInFlight)

why shared_ptr? none of the callers ever actually seem to share the memory...

Could you just return the PlayFabEventAPI object by value and rely on RVO and/or move semantics to avoid copies? If that fails, at the very least this seems like it should be unique_ptr

---
Refers to: source/test/TestApp/PlayFabEventTest.cpp:355 in 28a9118. [](commit_id = 28a911864c0819fb47cc78e81aca866a6ece7688, deletion_comment = False)

source/test/TestApp/PlayFabEventTest.cpp
```diff
@@ -286,6 +335,15 @@ namespace PlayFabUnit
     {
         eventTestContext = nullptr;
         eventApi = nullptr;
+
+        if (testThreadPool.size() > 0)
+        {
+            for (auto& thread : testThreadPool)
```
ScottMunroMS: >testThreadPool [](start = 32, length = 14)

I don't understand why this thread pool needs to be static state. Wouldn't it be clearer if the test which used this threadPool (which is currently only 1) just explicitly joined all of its threads before returning? seems like it makes for a simpler understanding.

Unnecessarily sharing memory statically often leads to problems and is best avoided.

source/test/TestApp/PlayFabEventTest.cpp
```diff
@@ -286,6 +335,15 @@ namespace PlayFabUnit
     {
         eventTestContext = nullptr;
         eventApi = nullptr;
+
+        if (testThreadPool.size() > 0)
+        {
```
ScottMunroMS: this size check is unnecessary. if the size is 0 then the loop won't execute and the clear will be a no-op

source/test/TestApp/PlayFabEventTest.h
```diff
@@ -57,11 +64,19 @@ namespace PlayFabUnit
             static int eventFailCount;
             static std::string eventFailLog;
 
+            static const unsigned int numThreads = 6;
+            static const unsigned int numEventsPerThread = 5;
+            static const unsigned int numTotalThreadedEvents = numThreads * numEventsPerThread;
+            static std::vector<std::thread> testThreadPool;
+            static unsigned int numEventsHeard;
```
ScottMunroMS: >numEventsHeard [](start = 32, length = 14)

nit: I'll leave it to your discretion but "heard" doesn't feel like a very precise way to describe what this value is measuring. "numEventsFired"? "numEventsSent"? "numEventsSentToPlaystream"?

source/test/TestApp/PlayFabEventTest.cpp
```diff
@@ -91,12 +91,23 @@ namespace PlayFabUnit
...
 
     void PlayFabEventTest::NonStaticEmitEventCallback(std::shared_ptr<const PlayFab::IPlayFabEvent> /*event*/, std::shared_ptr<const PlayFab::IPlayFabEmitEventResponse> /*response*/)
     {
         (*eventTestContext)->Pass("Private member called back!");
     }
 
+    void PlayFabEventTest::MyThreadingCallback(std::shared_ptr<const PlayFab::IPlayFabEvent> /*event*/, std::shared_ptr<const PlayFab::IPlayFabEmitEventResponse> /*response*/)
+    {
+        numEventsHeard++;
+        if(numEventsHeard == numTotalThreadedEvents)
```
ScottMunroMS: >numTotalThreadedEvents [](start = 29, length = 22)

only one test ever (intentionally) expects this exact number of events. The other two tests don't have any relation to this constant and only work because they are firing greater than or equal to this number of events. I would consider this another example where reliance on shared static state has led to less clear and semantically incorrect code. I think there's a better way to do what you are trying to do using condition variables. Let me see if I can jot down my idea in some quick pseudo because I'm having difficulty putting it in non-code words

source/test/TestApp/PlayFabEventTest.cpp
```diff
@@ -218,6 +229,40 @@ namespace PlayFabUnit
         std::bind(&PlayFabEventTest::NonStaticEmitEventCallback, this, std::placeholders::_1, std::placeholders::_2));
     }
 
+    void PlayFabEventTest::GenericMultiThreadedTest(unsigned int pNumThreads, unsigned int pNumEventsPerThread)
```
ScottMunroMS: >p [](start = 65, length = 1)

do you guys use hungarian? what does this 'p' prefix mean?

source/test/TestApp/PlayFabEventTest.h
ScottMunroMS: as pointed out in other comments I have some concerns about the use of shared static state.

I might recommend you do a larger sweep of your test code to eliminate as much shared static state as possible. The only things that I think should even be considered candidates for shared static state should be values that have to persist between tests (such as eventPassCount and eventFailCount) but there's likely a way to rework those to not be static as well.

source/test/TestApp/PlayFabEventTest.cpp
```diff
@@ -91,12 +91,23 @@ namespace PlayFabUnit
...
     {
         (*eventTestContext)->Pass("Private member called back!");
     }
 
+    void PlayFabEventTest::MyThreadingCallback(std::shared_ptr<const PlayFab::IPlayFabEvent> /*event*/, std::shared_ptr<const PlayFab::IPlayFabEmitEventResponse> /*response*/)
+    {
+        numEventsHeard++;
+        if(numEventsHeard == numTotalThreadedEvents)
+        {
+            (*eventTestContext)->Pass("Threaded callback worked as expected!");
```
ScottMunroMS: >Pass [](start = 33, length = 4)

what scenarios would lead to this test ever failing (not counting halting indefinitely)?

source/test/TestApp/PlayFabEventTest.cpp
ScottMunroMS: here's some pseudo-ish code

```
void GenericMultiThreadedTest(pNumThreads, pNumEventsPerThread)
{
    bool testDone = false;
    uint32_t totalEventsToWaitFor = pNumThreads * pNumEventsPerThread;

    auto EmitEventCallback = [<captured state>](<callbackParams>)
    {
        // Take special care to make sure multiple threads can't write this at the same time
        totalEventsToWaitFor -= 1;
        if (totalEventsToWaitFor == 0)
        {
            PassTest()!
        }
    }

    for (unsigned int thread = 0; thread < pNumThreads; ++thread)
    {
        testThreadPool.emplace_back(
            [pNumEventsPerThread, &EmitEventCallback]()
            {
                std::shared_ptr<PlayFabEventAPI*> api = SetupEventTest();
                for (unsigned int i = 0; i < pNumEventsPerThread; ++i)
                {
                    (*api)->EmitEvent(MakeEvent(0, PlayFabEventType::Default), EmitEventCallback);
                }
            });
    }

    {
        // Wait for all events to fire
        // TODO:
    }

    for (thread : testThreadPool)
    {
        thread.join();
    }
    testThreadPool.clear();
}
```

I didn't bother writing out any of the synchronization bits (figured since I wasn't writing the code and wouldn't bother verifying what I wrote those kinds of details would best be left as an exercise for you) but those are really important and must be done.

anyway I think this pseudo accomplishes removing a bunch of static state and making all of the relevant test code available in one Test function so you don't have to jump around the file to understand what happening

---
In reply to: [345442525](https://github.com/PlayFab/XPlatCoreTemplate/pull/44#discussion_r345442525) [](ancestors = 345442525)

source/test/TestApp/PlayFabEventTest.cpp
MichaelMackendMS: Or stick with emplace_back but skip the thread ctor as follows:

testThreadPool.emplace_back([pNumEventsPerThread, this]() { ... });

---
In reply to: [345433235](https://github.com/PlayFab/XPlatCoreTemplate/pull/44#discussion_r345433235) [](ancestors = 345433235)

source/test/TestApp/PlayFabEventTest.h
MichaelMackendMS: >const [](start = 19, length = 5)

it's a good habit to use constexpr where possible.

source/test/TestApp/PlayFabEventTest.cpp
toddbellMSFT: Only possible failure state is a time out when we didn't hit that expected limit. 

---
In reply to: [345446233](https://github.com/PlayFab/XPlatCoreTemplate/pull/44#discussion_r345446233) [](ancestors = 345446233)

source/test/TestApp/PlayFabEventTest.h
toddbellMSFT: That fuller sounding refactor sounds out of the scope of this particular pull request.

---
In reply to: [345444246](https://github.com/PlayFab/XPlatCoreTemplate/pull/44#discussion_r345444246) [](ancestors = 345444246)

source/test/TestApp/PlayFabEventTest.cpp
toddbellMSFT: parameter, this was meant to be distinctly different from the previous static name, so it's now just excessive. 

But as it's still a parameter, it still makes some sense (but admittedly it's not consistent throughout)

---
In reply to: [345443034](https://github.com/PlayFab/XPlatCoreTemplate/pull/44#discussion_r345443034) [](ancestors = 345443034)

toddbellMSFT:
>     std::shared_ptr<PlayFabEventAPI*> PlayFabEventTest::SetupEventTest(int maxBatchWaitTime, int maxItemsInBatch, int maxBatchesInFlight)

I agree it's concerning. I tried modifying it but the changes started to cascade beyond the scope of the threaded tests.

---
In reply to: [553106633](https://github.com/PlayFab/XPlatCoreTemplate/pull/44#issuecomment-553106633) [](ancestors = 553106633,553106455)

---
Refers to: source/test/TestApp/PlayFabEventTest.cpp:355 in 28a9118. [](commit_id = 28a911864c0819fb47cc78e81aca866a6ece7688, deletion_comment = False)

source/test/TestApp/PlayFabEventTest.cpp
ScottMunroMS: I don't think it makes sense if this style isn't consistent with any other method. it's just visual noise (and I don't think parameter is a typical prefix that is even used by those who use hungarian notation so it's double confusing. reactivating for consideration).

---
In reply to: [345516951](https://github.com/PlayFab/XPlatCoreTemplate/pull/44#discussion_r345516951) [](ancestors = 345516951,345443034)

source/test/TestApp/PlayFabEventTest.cpp
```diff
@@ -218,6 +220,56 @@ namespace PlayFabUnit
...
+                    (*api)->EmitEvent(MakeEvent(0, PlayFabEventType::Default),
+                    [&localNumEventsReturned, pNumEventsPerThread, localTotalEventsExpected]
+                    (std::shared_ptr<const PlayFab::IPlayFabEvent>, std::shared_ptr<const PlayFab::IPlayFabEmitEventResponse>)
+                    {
+                        localNumEventsReturned++;
+                        if (localNumEventsReturned == localTotalEventsExpected)
+                        {
+                            (*eventTestContext)->Pass("Threaded callback worked as expected!");
+                        }
+                    });
```
ScottMunroMS: nit: it's pretty hard to read this EmitEvent call with the lambda placed inline here. Recommend saving it is a named temporary somewhere (preferably outside of all of the loops since you don't need a unique lambda for each event)

```
auto EmitEventCallback = ...;
(*api)->EmitEvent(MakeEvent(0, Default, emitEventCallback));
```

source/test/TestApp/PlayFabEventTest.cpp
```diff
@@ -218,6 +220,56 @@ namespace PlayFabUnit
...
+                }
+                std::this_thread::yield();
+            });
```
ScottMunroMS: is this necessary?

source/test/TestApp/PlayFabEventTest.cpp
```diff
@@ -218,6 +220,56 @@ namespace PlayFabUnit
...
+        {
+            testThreadPool.emplace_back([&localNumEventsReturned, localTotalEventsExpected, pNumEventsPerThread, this]() {
+                std::shared_ptr<PlayFabEventAPI*> api = SetupEventTest();
+                for (uint32_t i = 0; i < pNumEventsPerThread; ++i)
+                {
+                    (*api)->EmitEvent(MakeEvent(0, PlayFabEventType::Default),
+                    [&localNumEventsReturned, pNumEventsPerThread, localTotalEventsExpected]
+                    (std::shared_ptr<const PlayFab::IPlayFabEvent>, std::shared_ptr<const PlayFab::IPlayFabEmitEventResponse>)
+                    {
+                        localNumEventsReturned++;
```
ScottMunroMS: >localNumEventsReturned [](start = 24, length = 22)

every thread has access to this value by reference. access to it needs to be properly synchronized to prevent concurrent writes/reads

ScottMunroMS:
>     std::shared_ptr<PlayFabEventAPI*> PlayFabEventTest::SetupEventTest(int maxBatchWaitTime, int maxItemsInBatch, int maxBatchesInFlight)

this is marked won't fix, but its usually nice to justify why it won't be fixed.

Is it not worth doing ever because of some reason? that'd be nice to know?

is it just out of scope for this change? then a new work item should be made to fix it later and ideally that would then be communicated back here so I know that is why this isn't being done in this change. It's totally reasonable to not want to allow a whole bunch of unrelated changes to get lumped into a PR but unless it's just a bad idea you usually still want tracking to avoid accruing technical debt

---
In reply to: [553107073](https://github.com/PlayFab/XPlatCoreTemplate/pull/44#issuecomment-553107073) [](ancestors = 553107073)

---
Refers to: source/test/TestApp/PlayFabEventTest.cpp:355 in 28a9118. [](commit_id = 28a911864c0819fb47cc78e81aca866a6ece7688, deletion_comment = False)

ScottMunroMS:
>     std::shared_ptr<PlayFabEventAPI*> PlayFabEventTest::SetupEventTest(int maxBatchWaitTime, int maxItemsInBatch, int maxBatchesInFlight)

It's fine to say its out of scope but this is a big enough hole in your tests that I think it's worth making a new work item and addressing very soon.

---
In reply to: [553181807](https://github.com/PlayFab/XPlatCoreTemplate/pull/44#issuecomment-553181807) [](ancestors = 553181807,553106633,553106455)

---
Refers to: source/test/TestApp/PlayFabEventTest.cpp:355 in 28a9118. [](commit_id = 28a911864c0819fb47cc78e81aca866a6ece7688, deletion_comment = False)

source/test/TestApp/PlayFabEventTest.h
ScottMunroMS: I agree that updating this file so that there is no shared static state is out of scope for this refactor, but I think even without doing that I don't think this thread pool needs to be static right?

---
In reply to: [345512283](https://github.com/PlayFab/XPlatCoreTemplate/pull/44#discussion_r345512283) [](ancestors = 345512283,345444246)

source/test/TestApp/PlayFabEventTest.cpp
ScottMunroMS: actually I did some more googling and I see some people doing  this in other places but not calling it hungarian (it's more like the 'm' prefix). that's fine i'll bite on this.

---
In reply to: [345524976](https://github.com/PlayFab/XPlatCoreTemplate/pull/44#discussion_r345524976) [](ancestors = 345524976,345516951,345443034)

ScottMunroMS:
marking approved but I would appreciate tracking items for those two test bugs I pointed out

toddbellMSFT:
>     std::shared_ptr<PlayFabEventAPI*> PlayFabEventTest::SetupEventTest(int maxBatchWaitTime, int maxItemsInBatch, int maxBatchesInFlight)

[https://dev.azure.com/playfab/PlayFab/_workitems/edit/38092](https://dev.azure.com/playfab/PlayFab/_workitems/edit/38092)

---
In reply to: [553192132](https://github.com/PlayFab/XPlatCoreTemplate/pull/44#issuecomment-553192132) [](ancestors = 553192132,553181807,553106633,553106455)

---
Refers to: source/test/TestApp/PlayFabEventTest.cpp:355 in 28a9118. [](commit_id = 28a911864c0819fb47cc78e81aca866a6ece7688, deletion_comment = False)

toddbellMSFT:
>     std::shared_ptr<PlayFabEventAPI*> PlayFabEventTest::SetupEventTest(int maxBatchWaitTime, int maxItemsInBatch, int maxBatchesInFlight)

The answer to this particular issue was resolved with the above issue. The memory semantics will need to be addressed in the noted work item. 

---
In reply to: [553191512](https://github.com/PlayFab/XPlatCoreTemplate/pull/44#issuecomment-553191512) [](ancestors = 553191512,553107073)

---
Refers to: source/test/TestApp/PlayFabEventTest.cpp:355 in 28a9118. [](commit_id = 28a911864c0819fb47cc78e81aca866a6ece7688, deletion_comment = False)

ScottMunroMS:
Thanks!

---
In reply to: [553522083](https://github.com/PlayFab/XPlatCoreTemplate/pull/44#issuecomment-553522083) [](ancestors = 553522083)

source/test/TestApp/PlayFabEventTest.cpp
```diff
@@ -210,14 +211,67 @@ namespace PlayFabUnit
...
+                {
+                    std::shared_ptr<PlayFabEventAPI*> api = SetupEventTest();
+                    for (uint32_t i = 0; i < pNumEventsPerThread; ++i)
+                    {
+                        (*api)->EmitEvent(MakeEvent(0, PlayFabEventType::Default),
+                            [&eventsRemaining, pNumEventsPerThread]
+                            (std::shared_ptr<const PlayFab::IPlayFabEvent>, std::shared_ptr<const PlayFab::IPlayFabEmitEventResponse>)
+                            {
+                                eventsRemaining--;
+                                if (eventsRemaining == 0)
```
ScottMunroMS: >eventsRemaining [](start = 36, length = 15)

I think what you had before was correct.
You want to do an atomic write-then-read

by splitting the `--` from the 0 check it's no longer atomic. Multiple separate threads might trip the if-check now

toddbellMSFT: this was a bad merge from a different branch. Thanks for catching this :)

---
In reply to: [346620630](https://github.com/PlayFab/XPlatCoreTemplate/pull/44#discussion_r346620630) [](ancestors = 346620630)

PR #45, opened 2019-11-13
Make FromJsonUtilT() predictable for time_t.

MSFT-Chris-Barrett:
mktime() and _mkgmtime() assume tm's tm_isdst is set to a deliberate,
known value.  When it's undefined-- and std::get_time() doesn't write to
it-- the result is unpredictable.  A zero value means "DST is not in
effect", positive means "DST is in effect", and a negative value means
"check the current locale/environment for whether or not DST is in
effect".  Adjustments _may_ be made to the resulting value depending on
this behavior.

Fixes https://github.com/PlayFab/XPlatCppSdk/issues/34

Change *should* be a safe minimal that just makes this function's result
predictable/consistent.  But don't consider this change to be
well-tested just yet.

Ideally, this function wouldn't ignore the incoming value's timezone, if
given.

### DIFF
--- a/source/code/include/playfab/PlayFabBaseModel.h
+++ b/source/code/include/playfab/PlayFabBaseModel.h
@@ -119,6 +119,7 @@ namespace PlayFab
         tm timeStruct = {};
         std::istringstream iss(timeStr);
         iss >> std::get_time(&timeStruct, "%Y-%m-%dT%T");
+        timeStruct.tm_isdst = 0;  // 0 means "not in DST".  FromJsonUtilT() assumes UTC/Zulu always.
 #if defined(PLAYFAB_PLATFORM_PLAYSTATION)
         output = mktime(&timeStruct);
 #elif defined(PLAYFAB_PLATFORM_IOS) || defined(PLAYFAB_PLATFORM_ANDROID) || defined(PLAYFAB_PLATFORM_LINUX)


pgilmorepf:
I'm actually of the opposite opinion.
I think this will break the time calculation when it is DST in your time zone, and the resulting time will be 1 hour off of actual for those people in those time zones.

MSFT-Chris-Barrett:
> 
> 
> I'm actually of the opposite opinion.
> I think this will break the time calculation when it is DST in your time zone, and the resulting time will be 1 hour off of actual for those people in those time zones.

As written, PlayFab will sometimes behave as this change describes, will sometimes behave as if DST is in effect (regardless of its actual status), and will sometimes check the current locale to determine its behavior.  Without some kind of change setting `tm_isdst` to a negative, zero, or positive value, the PlayFab SDK's behavior is not currently well-defined.  This is because the default-constructed timeStruct object never has `tm_isdst` written to.  So debug builds-- depending on the configuration-- will tend to have either 0 or a specific pattern written to it.  Most release/publish builds will tend to have whatever "garbage" data was on the stack.  Note the struct being used defines no default values, nor a default constructor.

Please see documentation on `std::get_time()`, such as the line `Note: tm_isdst is not written to, and needs to be set explicitly for use with functions such as mktime` from https://en.cppreference.com/w/cpp/io/manip/get_time .
Also see this documentation on the Windows-only `_mkgmtime()`, which describes it following similar behavior: https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/mkgmtime-mkgmtime32-mkgmtime64?view=vs-2019

PR #46, opened 2019-11-14
Exes and static libraries can now agree on the size of PlayFabApiSettings and PlayFabAuthenticationContext

pgilmorepf:
Resolves allocation/deallocation/segment-fault/out-of-bounds issues for XPlat, after the instance data isolation fixes
The fix itself, is to ensure that all fields always exist on these classes, regardless of defines, and only their access level changes when defines are changed.
Full design discussion and notes available in task 38104.

Other changes:
Separate the TestApp into TestClientApp and TestServerApp so that we can test both scenarios.
Add some allocations/deletes to the begining of the test-apps so they can verify these fixes work as expected.
Wrapping a lot of files in #ifndef DISABLE_PLAYFABCLIENT_API, since they were always client specific (but wasn't well tested before)
Making the logging funcs in TestApp public, so they can be called from the main funcs.

### DIFF
--- a/source/build/Windows/MainWin32.cpp
+++ b/source/build/Windows/MainWin32.cpp
@@ -3,11 +3,56 @@
 #include "TestAppPch.h"
 #include "TestApp.h"
 
+#include <playfab/PlayFabApiSettings.h>
+#include <playfab/PlayFabAuthenticationContext.h>
+
 // Win32 Entry Point
 int main()
 {
-    PlayFabUnit::TestApp testApp;
+    // Settings
+    PlayFab::PlayFabApiSettings* testPtrSettings = new PlayFab::PlayFabApiSettings();
+    testPtrSettings->titleId = "this should succeed";
+#if defined(ENABLE_PLAYFABSERVER_API) || defined(ENABLE_PLAYFABADMIN_API)
+    testPtrSettings->developerSecretKey = "this is conditional";
+#endif
+    delete testPtrSettings;
 
-    int result = testApp.Main();
-    return result;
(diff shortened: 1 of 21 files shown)


source/build/Windows/MainWin32.cpp
```diff
@@ -3,11 +3,56 @@
...
 #include "TestApp.h"
 
+#include <playfab/PlayFabApiSettings.h>
+#include <playfab/PlayFabAuthenticationContext.h>
+
 // Win32 Entry Point
 int main()
 {
-    PlayFabUnit::TestApp testApp;
+    // Settings
```
PushpadantK: >Settings [](start = 7, length = 8)

I don't think this block of testing belongs in main, you can move this as part of a different test where other tests are, this feels really out of place.

pgilmorepf: The main point of this here is simplicity.
This code has to live in the exe, rather than the lib.
All the other tests live in the lib.

I think the best I could do is implement a separate test suite in the exe, containing this code.
It would look like the other tests, but it would still live in the exe.

PushpadantK: even though its part of exe, can it be part of testapp, I think there should be a central place where all the test logic lives

---
In reply to: [346575631](https://github.com/PlayFab/XPlatCoreTemplate/pull/46#discussion_r346575631) [](ancestors = 346575631)

pgilmorepf: I agree.
I will address this in my timestamps branch, because it'll be my next checkin, and it is hard to go back and forth between them when they have overlapping changes.

pgilmorepf:
Running automated tests now.

pgilmorepf:
tests passed.

PR #47, opened 2019-11-15
Move platform-specific timestamp code to PlatformUtils

pgilmorepf:
Primary change is to move platform-specific code w/ accompanying platform #ifdef's to platform-specific files, so the consoles don't have to replace so much code in so many files.

Secondary:
* Move the allocation tests to their own test suite, and out of main() (Though, I fear doing so lessens their ability to actually detect the problem they're there to detect)
* requestGetParams wasn't supposed to be static, it was an incomplete portion of the api instance data isolation change.
* Discovered a serialization issue in the JSON Reader, throwing an exception, causing the program to hang. Wrapped it in a try/catch, which lets the program fail out gracefully.
* Found and removed some unused code
* Fixed some more warnings
* Converted a few hard-coded values to template-variables (same value for now)
* Lots of test code stops using time_t and switches to std::chrono
    * The public models still use time_t, but we're discovering these might be incorrect
* Test code will ouptut test results a little sooner, and more gracefully handle cloudscript test-submission failure
* Random bugfixes and code cleanup

Output for template changes available here:
https://github.com/PlayFab/XPlatCppSdk/pull/36/files


source/code/include/playfab/PlayFabPlatformUtils.h
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    inline std::string LocalTimeTToUtcString(time_t now)
+    {
+        tm timeInfo;
+#if defined(PLAYFAB_PLATFORM_WINDOWS) || defined(PLAYFAB_PLATFORM_XBOX)
+        gmtime_s(&timeInfo, &now);
+#elif defined(PLAYFAB_PLATFORM_LINUX) || defined(PLAYFAB_PLATFORM_IOS) || defined(PLAYFAB_PLATFORM_ANDROID) || defined(PLAYFAB_PLATFORM_PLAYSTATION)
+        timeInfo = *gmtime(&now);
+#endif
```
jdeweyMSFT: Should there be an #else clause that generates a compiler error? If not, this function will silently return uninitialized memory on new platforms.

source/code/include/playfab/PlayFabPlatformUtils.h
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    inline void AppendIntToString(int value, std::string& output)
+    {
+        // itoa is not available in android
+        char buffer[16];
+#if defined(PLAYFAB_PLATFORM_IOS) || defined(PLAYFAB_PLATFORM_ANDROID) || defined(PLAYFAB_PLATFORM_LINUX) || defined(PLAYFAB_PLATFORM_PLAYSTATION) || defined(PLAYFAB_PLATFORM_SWITCH)
+        sprintf(buffer, "%d", value);
+#else // PLAYFAB_PLATFORM_IOS || PLAYFAB_PLATFORM_ANDROID || PLAYFAB_PLATFORM_LINUX || PLAYFAB_PLATFORM_PLAYSTATION || PLAYFAB_PLATFORM_SWITCH
+        sprintf_s(buffer, "%d", value);
```
jdeweyMSFT: I would expect the use of _s variants to be opt-in per platform, rather than opt-out. So `#if defined(windows) || defined(xbox)`, use `sprintf_s`, `#else` use `sprintf`.

source/code/include/playfab/PlayFabPlatformUtils.h
```diff
@@ -0,0 +1,75 @@
...
+    inline TimePoint GetTimePointNow()
+    {
+        return std::chrono::system_clock::now();
+    }
+
+    inline time_t GetPlayFabTimeTNow()
+    {
+        time_t now = std::chrono::system_clock::to_time_t(GetTimePointNow());
+#if defined(PLAYFAB_PLATFORM_IOS) || defined(PLAYFAB_PLATFORM_ANDROID) || defined(PLAYFAB_PLATFORM_LINUX) || defined(PLAYFAB_PLATFORM_SWITCH)
+        return static_cast<Json::Int64>(now);
```
jdeweyMSFT: The function returns a time_t, so why is it being explicitly cast to Json::Int64 just to be implicitly cast back?

pgilmorepf: Scott had the same comment on the earlier failed PR.
I agree it's weird, and I don't know why.
I don't intend to change it in this PR, because this PR just moves this code from one file to another.
I'll file a bug about how it's weird for another PR.

source/code/include/playfab/PlayFabPlatformMacros.h
```diff
@@ -29,4 +29,24 @@
 // Hence to specify only Windows, we have check for ! _DURANGO.
 #if defined(_WIN32) && !defined(_DURANGO)
 #define PLAYFAB_PLATFORM_WINDOWS
-#endif //_WIN32
\ No newline at end of file
+#endif //_WIN32
+
```
jdeweyMSFT: This section is very different from the rest of the file, and the purpose stated in the file's header comment block. Consider creating a new PlayFabPlatformTypes.h

source/code/source/playfab/PlayFabCallRequestContainer.cpp
```diff
@@ -32,11 +32,16 @@ namespace PlayFab
...
-
-        if (parsingSuccessful)
+        try {
+            bool parsingSuccessful = reader->parse(reqBod, reqBod + reqBodLength, &request, &errs);
+            if (parsingSuccessful)
+            {
+                errorWrapper.Request = request;
+            }
+        }
+        catch (std::exception)
```
jdeweyMSFT: Prefer catching by const ref: `catch (const std::exception&)`

source/code/source/playfab/PlayFabCallRequestContainer.cpp
```diff
@@ -32,11 +32,16 @@ namespace PlayFab
...
         {
-            errorWrapper.Request = request;
+            // We can't parse the request back into a JSON::Value, so the caller won't receive it back...?
```
jdeweyMSFT: Did you mean for this to sound ambiguous?

source/code/include/playfab/PlayFabPlatformUtils.h
pgilmorepf: Bug Filed.

source/code/include/playfab/PlayFabPlatformUtils.h
pgilmorepf: I'd like to limit this PR to just moving existing, working code, from other locations to PlatformUtils.
Your comment seems reasonable, so I'll file a task.

pgilmorepf: Task filed

source/code/include/playfab/PlayFabPlatformUtils.h
pgilmorepf: task filed

source/code/source/playfab/PlayFabCallRequestContainer.cpp
pgilmorepf: Yes?
I tried to make it declare this call invalid/failure here, but that didn't work.
Essentially, it's a bug that we hadn't seen before, and resolving it properly wasn't trivial.
Good call though, I'll file a bug.

pgilmorepf: Bug filed.

source/code/include/playfab/PlayFabPlatformMacros.h
pgilmorepf: If I get a second vote agreeing with you, I'll do that.

source/test/TestApp/TestReport.cpp
```diff
@@ -26,12 +28,7 @@ namespace PlayFabUnit
         json["errors"] = errors;
         json["skipped"] = skipped;
         json["time"] = time;
-#if defined(PLAYFAB_PLATFORM_IOS) || defined(PLAYFAB_PLATFORM_ANDROID) || defined(PLAYFAB_PLATFORM_LINUX) || defined(PLAYFAB_PLATFORM_SWITCH)
-        json["timestamp"] = static_cast<Json::Int64>(std::chrono::system_clock::to_time_t(timeStamp));
-#else // PLAYFAB_PLATFORM_IOS || PLAYFAB_PLATFORM_ANDROID || PLAYFAB_PLATFORM_LINUX || PLAYFAB_PLATFORM_SWITCH
-        json["timestamp"] = std::chrono::system_clock::to_time_t(timeStamp);
-#endif // PLAYFAB_PLATFORM_IOS || PLAYFAB_PLATFORM_ANDROID || PLAYFAB_PLATFORM_LINUX || PLAYFAB_PLATFORM_SWITCH
-
```
toddbellMSFT: Awesome! I like that you encapsulated the platform out of the test layer here

source/test/TestApp/TestRunner.cpp
```diff
@@ -8,18 +8,16 @@
 #include "TestRunner.h"
 #include "TestUtils.h"
 
-using namespace PlayFab;
```
toddbellMSFT: nit: this seems like a readability regression to require adding PlayFab:: in most places (BUT, it makes sense if we want to differentiate the static GetTimePointNow from a potentially similarly named function here.

templates/PlayFab_InstanceAPI.h.ejs
```diff
@@ -16,25 +15,25 @@ namespace PlayFab
...
+        PlayFab<%- api.name %>InstanceAPI(std::shared_ptr<PlayFabApiSettings> apiSettings);
+        PlayFab<%- api.name %>InstanceAPI(std::shared_ptr<PlayFabAuthenticationContext> authenticationContext);
         PlayFab<%- api.name %>InstanceAPI(std::shared_ptr<PlayFabApiSettings> apiSettings, std::shared_ptr<PlayFabAuthenticationContext> authenticationContext);
 <% } else {
-%>        explicit PlayFab<%- api.name %>InstanceAPI(std::shared_ptr<PlayFabAuthenticationContext> authenticationContext);
+%>        PlayFab<%- api.name %>InstanceAPI(std::shared_ptr<PlayFabAuthenticationContext> authenticationContext);
         PlayFab<%- api.name %>InstanceAPI(std::shared_ptr<PlayFabApiSettings> apiSettings, std::shared_ptr<PlayFabAuthenticationContext> authenticationContext);
 <% } %>
-        ~PlayFab<%- api.name %>InstanceAPI();
+        ~PlayFab<%- api.name %>InstanceAPI() = default;
```
toddbellMSFT: >= default [](start = 45, length = 9)

+1 I like this change.

source/code/include/playfab/PlayFabApiSettings.h
```diff
@@ -18,7 +18,7 @@ namespace PlayFab
         std::string developerSecretKey; // Developer secret key. These keys can be used in development environments.
 
     public:
-        static const std::map<std::string, std::string> requestGetParams;
```
toddbellMSFT: >static const [](start = 8, length = 12)

could this possibly break users if they utilized requestGetParams?

source/test/TestApp/TestRunner.cpp
pgilmorepf: Basically, we've been moving towards a stricter more verbose usage in all cases because of name conflicts in the past.
When it's trivial, I remove the most obvious and worst potential conflictors.

source/code/include/playfab/PlayFabApiSettings.h
pgilmorepf: Yes and no.
I already moved it from static PlayFabSettings, to PlayFabApiSettings.
The intent was to remove the static then, but somehow I missed it.
So, THIS static removal isn't going to break players, but the previous checkin might.
Either way, there's no real reason for them to use it, since it only defines extra headers for the client to send PlayFab, and the only legit usage is internal atm.

source/code/include/playfab/PlayFabPlatformUtils.h
```diff
@@ -0,0 +1,75 @@
...
+#else // PLAYFAB_PLATFORM_IOS || PLAYFAB_PLATFORM_ANDROID || PLAYFAB_PLATFORM_LINUX || PLAYFAB_PLATFORM_SWITCH
+        return now;
+#endif // PLAYFAB_PLATFORM_IOS || PLAYFAB_PLATFORM_ANDROID || PLAYFAB_PLATFORM_LINUX || PLAYFAB_PLATFORM_SWITCH
+    }
+
+    inline void AppendIntToString(int value, std::string& output)
```
ScottMunroMS: >AppendIntToString [](start = 16, length = 17)

In the earlier iteration of this change I suggested replacing this with calls to std::to_string. You suggested that this wasn't supported on all platforms. What platforms don't support this? It's a std library function. as long as you have c++11 support (which is required by the playfab API for other reasons) you should be fine.

[https://en.cppreference.com/w/cpp/string/basic_string/to_string](https://en.cppreference.com/w/cpp/string/basic_string/to_string)

source/code/include/playfab/PlayFabPlatformUtils.h
```diff
@@ -0,0 +1,75 @@
...
+        char buffer[16];
+#if defined(PLAYFAB_PLATFORM_IOS) || defined(PLAYFAB_PLATFORM_ANDROID) || defined(PLAYFAB_PLATFORM_LINUX) || defined(PLAYFAB_PLATFORM_PLAYSTATION) || defined(PLAYFAB_PLATFORM_SWITCH)
+        sprintf(buffer, "%d", value);
+#else // PLAYFAB_PLATFORM_IOS || PLAYFAB_PLATFORM_ANDROID || PLAYFAB_PLATFORM_LINUX || PLAYFAB_PLATFORM_PLAYSTATION || PLAYFAB_PLATFORM_SWITCH
+        sprintf_s(buffer, "%d", value);
+#endif // PLAYFAB_PLATFORM_IOS || PLAYFAB_PLATFORM_ANDROID || PLAYFAB_PLATFORM_LINUX || PLAYFAB_PLATFORM_PLAYSTATION || PLAYFAB_PLATFORM_SWITCH
+        output.append(buffer);
+    }
+
+    inline std::string LocalTimeTToUtcString(time_t now)
```
ScottMunroMS: >LocalTime [](start = 23, length = 9)

In the previous iteration of this change we discussed how I don't think "Local" is the right term to use here.

You said in your experiments that it always represented a local time.

I'm curious what your experiments were... you said "In my past experiments, parsing and printing the output of time_t required shifting/correcting for the timezone, given the input string."

what were you printing? some stringified form of the time_t? the time_t itself as an int or something?

ScottMunroMS:
paul, a lot of comments you are responding with "filed a bug". That's fine but could you also follow up with a link for all of these bugs? If we need to dig through this PR later it will be nice to have that record here

source/code/include/playfab/PlayFabPlatformUtils.h
```diff
@@ -0,0 +1,75 @@
...
+    {
+        tm timeInfo;
+#if defined(PLAYFAB_PLATFORM_WINDOWS) || defined(PLAYFAB_PLATFORM_XBOX)
+        gmtime_s(&timeInfo, &now);
+#elif defined(PLAYFAB_PLATFORM_LINUX) || defined(PLAYFAB_PLATFORM_IOS) || defined(PLAYFAB_PLATFORM_ANDROID) || defined(PLAYFAB_PLATFORM_PLAYSTATION)
+        timeInfo = *gmtime(&now);
```
ScottMunroMS: >gmtime [](start = 20, length = 6)

in the earlier iteration I warned that this isn't threadsafe or reentrant. You said you'd file a bug. can you post the link to the bug here?

source/code/include/playfab/PlayFabPlatformUtils.h
pgilmorepf: I commented that std::to_string doesn't exist in all environments.

source/code/include/playfab/PlayFabPlatformUtils.h
```diff
@@ -0,0 +1,75 @@
...
+        sprintf_s(buffer, "%d", value);
+#endif // PLAYFAB_PLATFORM_IOS || PLAYFAB_PLATFORM_ANDROID || PLAYFAB_PLATFORM_LINUX || PLAYFAB_PLATFORM_PLAYSTATION || PLAYFAB_PLATFORM_SWITCH
+        output.append(buffer);
+    }
+
+    inline std::string LocalTimeTToUtcString(time_t now)
+    {
+        tm timeInfo;
+#if defined(PLAYFAB_PLATFORM_WINDOWS) || defined(PLAYFAB_PLATFORM_XBOX)
+        gmtime_s(&timeInfo, &now);
```
ScottMunroMS: >gmtime_s [](start = 8, length = 8)

in the last iteration, I said, gmtime_s returns an errno_t which should be handled or asserted. You filed a bug. Can you follow up with that bug link?

source/code/include/playfab/PlayFabPlatformUtils.h
pgilmorepf: https://github.com/PlayFab/XPlatCoreTemplate/blob/pg/timestamp/source/test/TestApp/PlayFabApiTest.cpp#L287

I can't say exactly what I did/tested ages ago.
The test above is the current test, and satisfying that test should be sufficient.

That said, I'm hours deep into testing this change, so I'm very hesitant to make changes to this PR. I'm happy to file a task for a non-breaking redesign/rename/fix.

pgilmorepf:
Filed a bug and links:
I'd prefer not to. These comments are public facing.

source/code/include/playfab/PlayFabPlatformUtils.h
pgilmorepf: Public facing comments.

source/code/include/playfab/PlayFabPlatformUtils.h
pgilmorepf: Public facing comments.

ScottMunroMS: >gmtime_s [](start = 8, length = 8)

the last time I brought up asserting this possible error return you responded by saying that it'd be a "whole new feature" to assert this return's success...

I don't think I understand. the c and c++ std have utilities for performing asserts. I would not considering adding a single line a new feature

```
errno_t err = gmtime_s(...);
std::assert(err == 0);
```

but maybe I misunderstood. what did you mean?

pgilmorepf: Yes, but we haven't discussed the ramifications of throwing asserts in code that customers call.

source/code/include/playfab/PlayFabPlatformMacros.h
ScottMunroMS: I agree with jeremy

source/code/include/playfab/PlayFabPlatformMacros.h
```diff
@@ -29,4 +29,24 @@
...
+#if defined(PLAYFAB_PLATFORM_WINDOWS) || defined(PLAYFAB_PLATFORM_XBOX)
+typedef signed __int64 Int64;
+typedef signed __int32 Int32;
```
ScottMunroMS: in a previous version I mentioned how redefining the fixed-width numeric types (e.g. int32_t) provided by the std library is unnecesary.

You mentioned you'd be willing to improve it in a separate change. Can you post the link to the work item for that implied punted work?

source/code/include/playfab/PlayFabPlatformUtils.h
ScottMunroMS: you already have asserts in customer code. check out PlayFabBaseModel.h line 25
`inline operator const BoxedType& () const { assert(notNull()); return *operator->(); }`

source/test/TestApp/TestRunner.cpp
ScottMunroMS: I followed up offline about this but please be careful of this. There are dependencies on these `using` statements in some places that would be painful to deal with if this style was adopted more broadly in the SDK

source/code/include/playfab/PlayFabPlatformUtils.h
ScottMunroMS: what environments/platforms? I'm arguing that it has to because it's part of the std library that you already depend on and you haven't seemed to provide a counter example that disproves that

source/code/include/playfab/PlayFabPlatformUtils.h
```diff
@@ -0,0 +1,70 @@
...
+    typedef std::chrono::time_point<std::chrono::system_clock> TimePoint;
+
+    inline TimePoint GetTimePointNow()
+    {
+        return std::chrono::system_clock::now();
+    }
+
+    inline time_t GetPlayFabTimeTNow()
+    {
+        return std::chrono::system_clock::to_time_t(GetTimePointNow());
```
ScottMunroMS: >std::chrono::system_clock::to_time_t(GetTimePointNow()); [](start = 15, length = 56)

nit: alternatively you could just `return time(0);`

source/code/include/playfab/PlayFabPlatformTypes.h
```diff
@@ -0,0 +1,21 @@
+// Copyright (C) Microsoft Corporation. All rights reserved.
+
+#if defined(PLAYFAB_PLATFORM_WINDOWS) || defined(PLAYFAB_PLATFORM_XBOX)
```
ScottMunroMS: I think this needs a `pragma once`

source/code/include/playfab/PlayFabPlatformMacros.h
pgilmorepf: If you have a suggestion for a better overall solution here, please let me know.
For now I'm just copy-pasting what's already there, into an appropriate file.

source/code/include/playfab/PlayFabPlatformUtils.h
pgilmorepf: I can't argue with your logic. It's possible that's been resolved. But we don't have automated testing across everything right now, so I'm trying to limit myself to the safest and most relevant changes. This function exists, and it has been tested to work on all platforms. std::to_string definitely didn't exist on at least one platform in the past, which forced us to make this function.
Out of scope for this PR.
Task filed.

source/code/include/playfab/PlayFabPlatformMacros.h
pgilmorepf: Done.

source/test/TestApp/TestApp.cpp
```diff
@@ -89,17 +96,21 @@ namespace PlayFabUnit
             std::bind(&TestApp::OnPostReportError, this, std::placeholders::_1, std::placeholders::_2),
             &testRunner.suiteTestReport);
 
-        while (cloudResponse.empty())
+        for (int i = 0; i < 300; i++)
```
toddbellMSFT: Nit: can the 300 be a named variable?

source/code/include/playfab/PlayFabJsonHeaders.h
```diff
@@ -6,7 +6,10 @@
 
 #include <playfab/PlayFabPlatformMacros.h>
 
-#ifdef PLAYFAB_PLATFORM_LINUX
+#if defined(PLAYFAB_PLATFORM_SWITCH)
+#include <json/json.h>
```
toddbellMSFT: including this particular json.h file gave me issues before about a monolithic header file and that we should really only use the sub classes (like json/value or json/reader) to save on space.

source/code/include/playfab/PlayFabCallRequestContainer.h
```diff
@@ -24,13 +24,16 @@ namespace PlayFab
         std::string GetFullUrl() const;
         std::shared_ptr<PlayFabApiSettings> GetApiSettings() const;
         std::shared_ptr<PlayFabAuthenticationContext> GetContext() const;
+        std::string GetRequestId() const;
+        void SetRequestId(const std::string& newRequestId);
```
ScottMunroMS: is this the x-requestid header that's included in the api response?

ScottMunroMS: I suppose I'm asking because I'm wondering if bumblelion's custom http plugin should be interacting with this at all? is this used internally in the SDK anywhere that would require a custom plugin to have set this?

---
In reply to: [348113780](https://github.com/PlayFab/XPlatCoreTemplate/pull/47#discussion_r348113780) [](ancestors = 348113780)

source/code/source/playfab/PlayFabEvent.cpp
```diff
@@ -89,6 +91,8 @@ namespace PlayFab
         this->eventContents.Payload[name] = value;
 #elif defined(PLAYFAB_PLATFORM_ANDROID)
         this->eventContents.Payload[name] = (Uint64)value;
+#elif defined(JSON_HAS_INT64)
```
ScottMunroMS: >JSON_HAS_INT64 [](start = 14, length = 14)

this elif feels super weird

the other guards are based on platform but this one is  feature flag? what happens if a platform like playstation or android also has this defined? it will overwrite the payload twice and you may not get what you expect.

It also doesn't make a lot of sense to me why these ifdef guards are event necessary in the first place. Why can't all platforms be set to the raw u64 value instead of 3 of them requiring casts?

source/test/TestApp/TestApp.cpp
```diff
@@ -89,17 +103,21 @@ namespace PlayFabUnit
             std::bind(&TestApp::OnPostReportError, this, std::placeholders::_1, std::placeholders::_2),
             &testRunner.suiteTestReport);
 
-        while (cloudResponse.empty())
+        for (int i = 0; i < CLOUDSCRIPT_TIMEOUT_MS; i+= CLOUDSCRIPT_TIMEOUT_INCREMENT)
```
ScottMunroMS: >+= CLOUDSCRIPT_TIMEOUT_INCREMENT [](start = 53, length = 32)

it may help clarity to place a compile-time assert that CLOUDSCRIPT_TIMEOUT_MS is evenly divisible by CLOUDSCRIPT_TIMEOUT_INCREMENT

`std::static_assert(CLOUDSCRIPT_TIMEOUT_MS  % CLOUDSCRIPT_TIMEOUT_INCREMENT == 0)`

note: this should not be impacted by your larger "how to do asserts" work item since this is a compile time assert and should not affect customers at runtime (also it's just in the test app)

source/test/TestApp/TestApp.cpp
```diff
@@ -89,17 +103,21 @@ namespace PlayFabUnit
...
         {
-            std::this_thread::sleep_for(TimeValueMs(100));
+            std::this_thread::sleep_for(std::chrono::milliseconds(CLOUDSCRIPT_TIMEOUT_INCREMENT));
+            if (!cloudResponse.empty())
```
ScottMunroMS: sleep_for in a loop + cloudResponse.empty() check looks like an overly verbose way to implement a condition_variable (I'd say your current code is anti pattern that enables concurrency bugs)

recommend looking at std::condition_variable specifically std::condition_variable::std::wait_for

it will let you wait for a resource to become available (cloudResponse) and will avoid the need to declare the CLOUDSCRIPT_TIMEOUT_INCREMENT  variable

source/code/include/playfab/PlayFabJsonHeaders.h
pgilmorepf: "monolithic header file" is a Unreal warning, not an XPlat one.

source/code/include/playfab/PlayFabCallRequestContainer.h
pgilmorepf: Yes, this is the PlayFab X-RequestID.
It's in the PlayFab specific CallRequestContainer, rather than the agnostic BaseCallRequestContainer, so this is the correct segregation.
As far as the HTTP layers go, yes, there's a lot of cross-contamination of PlayFab in those, and I agree it's a problem. But it's not anywhere near related to this PR.

pgilmorepf: These changes are a consequence of the separate independent console repos, which had this change as a console-specific element, but we're moving certain interface pieces to core to reduce the complexity of the console-specific over-writes.

source/test/TestApp/TestApp.cpp
pgilmorepf: Added a comment to the assert task.

source/test/TestApp/TestApp.cpp
pgilmorepf: I have a bug about an occasional TestApp deadlock issue, and your suggestion resonates with that bug.
I'm going to copy your comments into that bug.

source/code/source/playfab/PlayFabEvent.cpp
pgilmorepf: I agree.
I'll file a follow-up bug for this, because I'm rather desperate to get this checked in, as it's tested and working.

source/code/include/playfab/PlayFabPlatformUtils.h
pgilmorepf: While this PR gets all the projects to compile and tests to pass for the new PlatformUtils.h file, I agree the PlatformUtils.h file itself isn't done.
bug filed.

source/code/source/playfab/PlayFabCallRequestContainer.cpp
pgilmorepf: fixed

PR #48, opened 2019-11-16
Update PlayFabEventTest.cpp

toddbellMSFT:
Android and iOS deleted the copy constructor for unsigned ints on std::atomic. We get around that now by not making a copy, but just calling the always valid constructor!


toddbellMSFT:
https://pf-jenkins.playfab.com/view/SDK/job/SDK_Work_XPlat_Android/513/console
https://pf-jenkins.playfab.com/view/SDK/job/SDK_Work_XPlat_iOS/577/console

PR #50, opened 2019-11-20

toddbellMSFT:
this move can potentially be a problem if we have an error since we still call out this object after this std::move is called.


source/code/source/playfab/PlayFabWinHttpPlugin.cpp
```diff
@@ -212,7 +212,7 @@ namespace PlayFab
                         if (!GetBinaryPayload(reqContainer, payload, payloadSize))
                         {
                             // set string payload if binary wasn't provided
-                            requestBody = std::move(reqContainer.GetRequestBody());
+                            requestBody = reqContainer.GetRequestBody();
```
ScottMunroMS: this is a good change to make, but I don't totally understand your PR description. It sounds like you are implying the std::move might move the member data out of the reqContainer leaving the value inside of the reqContainer empty or invalid. This is not the case. GetRequestBody returns a std::string by value which invokes a copy. Therefore the std::move is only moving a copy and the original requestbody string is not modified regardless of whether you do the std::move or not.

the only thing that removing the std::move does is make this line more clear about what the actual return type it (and set it up for potential RVO goodness in some future edit where the declaration and initialization are on the same line)

PR #53, opened 2019-11-23
Update PlayFabEventPipeline.cpp

toddbellMSFT:
Adding a PlayFabError to IntakeEvent failing to take an event.

### DIFF
--- a/source/code/source/playfab/PlayFabEventPipeline.cpp
+++ b/source/code/source/playfab/PlayFabEventPipeline.cpp
@@ -114,14 +114,13 @@ namespace PlayFab
 
             emitEventError->ErrorName = "PlayFabEventPipeline IntakeEvent Error";
             emitEventError->ErrorMessage = "PlayFabEventPipeline did not accept the event. Please see ErrorDetails for more information.";
-            emitEventError->ErrorCode = PlayFabErrorCode::PlayFabErrorEventNotFound;
 
             emitEventError->HttpCode = 0;
             emitEventError->HttpStatus = "None";
 
             if(emitResult == EmitEventResult::Overflow)
             {
-                emitEventError->ErrorDetails = "PlayFabEventPipeline was unable to take the event due to memory limits. Please wait for batching to complete before retrying or increase the pipeline size.";
+                emitEventError->ErrorDetails = "PlayFabEventPipeline was unable to take the event due to memory limits. Please wait for batching to complete before retrying or increase the PlayFabEventBuffer size (see its constructor)";
             }
             else if (emitResult == EmitEventResult::Disabled)
             {
--- a/source/code/source/playfab/PlayFabEventPipeline.cpp
+++ b/source/code/source/playfab/PlayFabEventPipeline.cpp
@@ -110,6 +110,30 @@ namespace PlayFab
             auto playFabEmitEventResponse = std::make_shared<PlayFabEmitEventResponse>();
             playFabEmitEventResponse->emitEventResult = emitResult;
 
+            std::shared_ptr<PlayFabError> emitEventError  = std::make_shared<PlayFabError>();
+
+            emitEventError->ErrorName = "PlayFabEventPipeline IntakeEvent Error";
+            emitEventError->ErrorMessage = "PlayFabEventPipeline did not accept the event. Please see ErrorDetails for more information.";
+            emitEventError->ErrorCode = PlayFabErrorCode::PlayFabErrorEventNotFound;
+
+            emitEventError->HttpCode = 0;
+            emitEventError->HttpStatus = "None";
+
+            if(emitResult == EmitEventResult::Overflow)
+            {
+                emitEventError->ErrorDetails = "PlayFabEventPipeline was unable to take the event due to memory limits. Please wait for batching to complete before retrying or increase the pipeline size.";
+            }
+            else if (emitResult == EmitEventResult::Disabled)
+            {
+                emitEventError->ErrorDetails = "PlayFabEventPipeline was unable to take the event due to being disabled. Please enable the pipeline before attempting to add events.";
+            }
+            else
+            {
+                emitEventError->ErrorDetails = "PlayFabEventPipeline was unable to take the event due to an unknown reason.";
+            }
+
+            playFabEmitEventResponse->playFabError = emitEventError;
+
             // call an emit event callback
             CallbackRequest(playFabEmitRequest, std::move(playFabEmitEventResponse));
         }


source/code/source/playfab/PlayFabEventPipeline.cpp
```diff
@@ -110,6 +110,30 @@ namespace PlayFab
...
+            emitEventError->ErrorName = "PlayFabEventPipeline IntakeEvent Error";
+            emitEventError->ErrorMessage = "PlayFabEventPipeline did not accept the event. Please see ErrorDetails for more information.";
+            emitEventError->ErrorCode = PlayFabErrorCode::PlayFabErrorEventNotFound;
+
+            emitEventError->HttpCode = 0;
+            emitEventError->HttpStatus = "None";
+
+            if(emitResult == EmitEventResult::Overflow)
+            {
+                emitEventError->ErrorDetails = "PlayFabEventPipeline was unable to take the event due to memory limits. Please wait for batching to complete before retrying or increase the pipeline size.";
```
pgilmorepf: (How does a customer increase the pipeline size?)

source/code/source/playfab/PlayFabEventPipeline.cpp
```diff
@@ -110,6 +110,30 @@ namespace PlayFab
...
+            }
+            else if (emitResult == EmitEventResult::Disabled)
+            {
+                emitEventError->ErrorDetails = "PlayFabEventPipeline was unable to take the event due to being disabled. Please enable the pipeline before attempting to add events.";
```
pgilmorepf: (How does a customer enable the pipeline?)

pgilmorepf:
I like when the error comment has more concrete code-suggestions about how to accomplish the task.
I'd provide an exact reference to the variable like EventThingy::enabled = true; or something.
Otherwise, it looks good.

source/code/source/playfab/PlayFabEventPipeline.cpp
```diff
@@ -110,6 +110,30 @@ namespace PlayFab
...
+            emitEventError->ErrorCode = PlayFabErrorCode::PlayFabErrorEventNotFound;
+
+            emitEventError->HttpCode = 0;
+            emitEventError->HttpStatus = "None";
+
+            if(emitResult == EmitEventResult::Overflow)
+            {
+                emitEventError->ErrorDetails = "PlayFabEventPipeline was unable to take the event due to memory limits. Please wait for batching to complete before retrying or increase the pipeline size.";
+            }
+            else if (emitResult == EmitEventResult::Disabled)
```
ScottMunroMS: >EmitEventResult::Disabled [](start = 35, length = 25)

is this result even possible anymore? I went looking through the source code for PlayFabEventBuffer and it doesn't look like there's actually any code which flips the `disabled` flag necessary to trigger this result

source/code/source/playfab/PlayFabEventPipeline.cpp
```diff
@@ -110,6 +110,30 @@ namespace PlayFab
...
+            std::shared_ptr<PlayFabError> emitEventError  = std::make_shared<PlayFabError>();
+
+            emitEventError->ErrorName = "PlayFabEventPipeline IntakeEvent Error";
+            emitEventError->ErrorMessage = "PlayFabEventPipeline did not accept the event. Please see ErrorDetails for more information.";
+            emitEventError->ErrorCode = PlayFabErrorCode::PlayFabErrorEventNotFound;
+
+            emitEventError->HttpCode = 0;
+            emitEventError->HttpStatus = "None";
+
+            if(emitResult == EmitEventResult::Overflow)
```
ScottMunroMS: >if [](start = 12, length = 2)

nit: instead of adding a new sequence of if-else statements, why not use the switch case above which sets `emitResult`

source/code/source/playfab/PlayFabEventPipeline.cpp
```diff
@@ -110,6 +110,30 @@ namespace PlayFab
             auto playFabEmitEventResponse = std::make_shared<PlayFabEmitEventResponse>();
             playFabEmitEventResponse->emitEventResult = emitResult;
 
+            std::shared_ptr<PlayFabError> emitEventError  = std::make_shared<PlayFabError>();
+
+            emitEventError->ErrorName = "PlayFabEventPipeline IntakeEvent Error";
+            emitEventError->ErrorMessage = "PlayFabEventPipeline did not accept the event. Please see ErrorDetails for more information.";
+            emitEventError->ErrorCode = PlayFabErrorCode::PlayFabErrorEventNotFound;
```
ScottMunroMS: >PlayFabErrorEventNotFound [](start = 58, length = 25)

this seems like a misleading error code. what does it mean that the event wasn't found? found from what? that doesn't seem like an accurate reflection of the problem which was that the event couldn't be placed in the event buffer

source/code/source/playfab/PlayFabEventPipeline.cpp
```diff
@@ -110,6 +110,30 @@ namespace PlayFab
             auto playFabEmitEventResponse = std::make_shared<PlayFabEmitEventResponse>();
             playFabEmitEventResponse->emitEventResult = emitResult;
 
+            std::shared_ptr<PlayFabError> emitEventError  = std::make_shared<PlayFabError>();
+
+            emitEventError->ErrorName = "PlayFabEventPipeline IntakeEvent Error";
+            emitEventError->ErrorMessage = "PlayFabEventPipeline did not accept the event. Please see ErrorDetails for more information.";
```
ScottMunroMS: >"PlayFabEventPipeline did not accept the event. Please see ErrorDetails for more information."; [](start = 43, length = 95)

what's the point of this string if it just redirects you to ErrorDetails? why not just put the ErrorDetails string in both ErrorMessage and ErrorDetails? maybe there's some distinction between ErrorDetails and ErrorMessage I'm not seeing yet, but ErrorMessage seems pretty unhelpful here.

source/code/source/playfab/PlayFabEventPipeline.cpp
toddbellMSFT: That will require moving the creation of the PlayFabError to before the previous switch statement. I originally thought that would look nicer too, but I think there are going to be more non error causing cases of this IntakeEvent call, which an empty event construction would be unnecessary for the most common expected case.

source/code/source/playfab/PlayFabEventPipeline.cpp
toddbellMSFT: I can agree that this error code is misleading. 

But there really is not a good error code for this particular issue. To add that error code, we will need to make a pf-main request to add that error code to all other api's as well (like java, unity and php) which won't have an event pipeline, adding noise to our already very long error code list. Would we want to leave this errorcode blank and expect users to read the ErrorName? Would a -1 casted PlayFab error code be acceptable (or even possible if it's an enum which I thought would be an unsigned int)

source/code/source/playfab/PlayFabEventPipeline.cpp
ScottMunroMS: I don't think it would. The only thing that needs to be declared before the switch is a variable to hold the string literal.
e.g.
```
const char* errorDetail = "";
switch (this->buffer.TryPut(request))
{
... // set errorDetail in switch
}

auto emitEventError = std::make_share<>()
// ...
emitEventError->ErrorDetails = errorDetail;
```

that being said this is all personal preference so i'll leave it to your discretion

source/code/source/playfab/PlayFabEventPipeline.cpp
toddbellMSFT: The distinction comes into play when a user calls the GenerateErrorReport(). But I guess we can leave out directing users to the details then since it will just show up.

source/code/source/playfab/PlayFabEventPipeline.cpp
toddbellMSFT: I noticed that as well. There's no way to really toggle that disabled flag on the Pipeline at the moment.

PR #54, opened 2019-11-25
Toddbell/plugin fix

toddbellMSFT:
PlayFab Party reported PlayFab was making 2 additional std::string allocations within the PluginManager. The below fix removes the use of a map in favor of a lock protected vector.


ScottMunroMS:
your description mentions you were doing O(N) lookup before anyway. I don't understand that. You were using std::map so wasn't the lookup O(LogN)? I don't think the difference here between O(N) vs O(LogN) is meaningful considering my previous points that you can only ever set one plugin in here anyway, but I figrured that'd be worth drilling into

ScottMunroMS:
please update your description to not use PlayFab Party's internal codename since this repo is public

source/code/source/playfab/PlayFabPluginManager.cpp
```diff
@@ -27,47 +27,43 @@ namespace PlayFab
...
-            plugins.erase(key);
-            plugins.insert({ key, std::move(plugin) });
+            if (pluginEntry.contract == contract &&
+                pluginEntry.name == instanceName)
+            {
+                return pluginEntry;
+            }
         }
+        plugins.push_back({contract, instanceName, nullptr});
+        return plugins.back();
```
ScottMunroMS: >back [](start = 23, length = 4)

I didn't realize this in my original proposal, but there's nothing prevent concurrent writes/reads to this vector. I think this function needs to mutually exclude access to this vector to be thread-safe. This problem also existed with the previous map implementation I believe

ScottMunroMS:
leaving a placeholder comment to figure out why the static std::string didn't work for you

source/code/source/playfab/PlayFabPluginManager.cpp
```diff
@@ -27,47 +29,43 @@ namespace PlayFab
...
         {
-            plugins.erase(key);
-            plugins.insert({ key, std::move(plugin) });
+            if (pluginEntry.contract == contract &&
+                pluginEntry.name == instanceName)
+            {
+                return pluginEntry;
+            }
         }
+        plugins.push_back({contract, instanceName, nullptr});
```
MichaelMackendMS: This could be refactored towards emplace_back to save a potential string allocation, but it's not a blocker given the use-case.

ScottMunroMS: I don't think emplace_back saves a string allocation.
Given move semantics, I think the only difference between empalce vs push_back is where the string allocation takes place. with push_back, the copy/allocation takes place when the r-value `{contract, instanceName, nullptr}` is constructed. After, the copied string value is moved into the constructed element. with emplace_back, the copy just happens as part of building the constructed element.
That being said, if emplace_back works here, I think it's the better API to use since it's more clear

source/code/source/playfab/PlayFabPluginManager.cpp
```diff
@@ -27,47 +29,46 @@ namespace PlayFab
...
+            std::unique_lock<std::mutex> lock(pluginsMutex);
+            for (PluginEntry& pluginEntry : plugins)
+            {
+                if (pluginEntry.contract == contract &&
+                    pluginEntry.name == instanceName)
+                {
+                    return pluginEntry;
+                }
+            }
+            plugins.emplace_back<PluginEntry>({contract, instanceName, nullptr});
```
ScottMunroMS: >{ [](start = 46, length = 1)

if you use emplace_back you should drop these braces (otherwise it's doing the exact same thing as push_back.

the plugins.emplace_back call also shouldn't require you to specify PluginEntry in the template parameter. I imagine that was only needed because you had the extraneous braces

toddbellMSFT: This actually doesn't seem to compile without the template syntax? Removing the template and curly brackets with just the parameters doesn't seem to map to the struct. we could make a struct outside the parentheses but that would incur similar excess copy issues.

PR #58, opened 2019-12-03
Fixing GetUrl for Android

pgilmorepf:
Discovered that the Android build for XPlat was non-functional, and that the automated tests somehow didn't catch it.

Primary Fix: GetUrl was called via the older mechanism, pre data isolation fix, which wasn't updated correctly in Android.
Secondary Fix: Fixing a weird segfault where jstring is apparently the incorrect type, despite documentation that it should be correct (temporarily fixed with "jstring" -> "auto" (Bug filed to investigate deeper later)
Ancillary Refactor: Updating the project to the latest Android Studio (somewhat incidental, as I had to install fresh, and it was automatic)


source/build/Android/build.gradle
```diff
@@ -7,7 +7,7 @@ buildscript {
         
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:3.4.0'
+        classpath 'com.android.tools.build:gradle:3.5.2'
```
toddbellMSFT: Feel free to won't fix. Version numbers seem like they could be inputs through a .ejs file. 

But it effectively would just move this number into another place unless we could make a call to get the latest version of gradle (or whatever is currently on the machine). 

While looking for Gradle environment variables I ran across: https://docs.gradle.org/current/userguide/single_versions.html which said we could possibly declare a version number range instead. BUT the very bottom of the page mentions we can declare a dependency constraint: 

A recommended practice for larger projects is to declare dependencies without versions and use dependency constraints for version declaration. The advantage is that dependency constraints allow you to manage versions of all dependencies, including transitive ones, in one place.

source/code/source/playfab/PlayFabAndroidHttpPlugin.cpp
```diff
@@ -424,7 +424,7 @@ namespace PlayFab
                 return false;
             }
 
-            jstring urlJstr = jniEnv->NewStringUTF("POST");
+            auto urlJstr = jniEnv->NewStringUTF("POST");
```
toddbellMSFT: This doesn't meet the requirements for auto use. I don't know the return type of NewStringUTF to be a jstring.

source/code/source/playfab/PlayFabAndroidHttpPlugin.cpp
```diff
@@ -435,15 +435,16 @@ namespace PlayFab
...
+            const std::string requestUrl = pfReqContainer->GetFullUrl();
 
             jmethodID methodId = jniEnv->GetMethodID(GetHelper().GetHttpRequestClass(), "setUrl", "(Ljava/lang/String;)Z");
             if (methodId == nullptr)
             {
                 return false;
             }
 
-            jstring urlJstr = jniEnv->NewStringUTF(requestUrl.c_str());
+            auto urlJstr = jniEnv->NewStringUTF(requestUrl.c_str());
```
toddbellMSFT: This doesn't meet the requirements for auto use. I don't know the return type of NewStringUTF to be a jstring.

source/code/source/playfab/PlayFabAndroidHttpPlugin.cpp
```diff
@@ -635,12 +628,12 @@ namespace PlayFab
             return ;
         }
 
-        jstring nameJstr = jniEnv->NewStringUTF(name);
+        auto nameJstr = jniEnv->NewStringUTF(name);
```
toddbellMSFT: same goes for the rest of the file

source/code/source/playfab/PlayFabAndroidHttpPlugin.cpp
```diff
@@ -435,15 +435,16 @@ namespace PlayFab
 
         // Call SetUrl
         {
-            const std::string requestUrl = GetUrl(requestTask);
+            CallRequestContainer* pfReqContainer = static_cast<CallRequestContainer*>(requestTask.requestContainer.get());
```
toddbellMSFT: It looks like pfReqContainer is only used to retrieve the GetFullUrl, i'd be a bit worried about any excess copies/casts if all we are doing is getting an internal member. Can we roll this call into the below line? 

caveat: it seems if the reqContainer doesn't get Initialize called (which I think we may be relying on a caller to do beforehand), theoretically .get() could return a nullptr which we may want to at least check for either a.) in the get or b.) right here (and any other get call)

source/code/source/playfab/PlayFabAndroidHttpPlugin.cpp
```diff
@@ -635,12 +628,12 @@ namespace PlayFab
...
         }
 
-        jstring nameJstr = jniEnv->NewStringUTF(name);
+        auto nameJstr = jniEnv->NewStringUTF(name);
         if (nameJstr == nullptr)
         {
             return ;
         }
-        jstring valueJstr = jniEnv->NewStringUTF(value);
+        auto valueJstr = jniEnv->NewStringUTF(value);
```
toddbellMSFT: auto rules don't apply to the lines in this file. It is not obvious what the return type is for NewStringUTF.

source/build/Android/build.gradle
toddbellMSFT: Somehow my comment didn't get posted? 

You should check out this Gradle doc page: https://docs.gradle.org/current/userguide/single_versions.html 

There's suggestions for larger projects to use "upgrade-able" ranges for versions numbers OR we could use constraints and wouldn't need to explicitly label the version number itself.

source/build/Android/gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,6 @@
-#Thu Apr 04 21:15:54 PDT 2019
+#Mon Dec 02 14:37:52 PST 2019
```
toddbellMSFT: nit: this will change whenever someone has to update the Android portion of the xplat sdk. Should this file be added to the .gitignore file? Or would this file not be properly generated if we open the Android Studio test project?

source/code/source/playfab/PlayFabAndroidHttpPlugin.cpp
toddbellMSFT: My previous comments weren't showing up on the review. I'm closing this one as it's an accidental duplicate.

source/build/Android/build.gradle
toddbellMSFT: My previous comments weren't showing up on the review. I'm closing this one as it's an accidental duplicate.

pgilmorepf: While I agree with you on principle, the fact of the matter is that it's already a fixed version, and my Android Studio just updated the project somewhat against my will.
Rolling it back may not be safe, given the other changes.
And the rest of what you say is out-of-scope for this bugfix.

source/code/source/playfab/PlayFabAndroidHttpPlugin.cpp
pgilmorepf: I know for certain that declaring the type to be jstring will compile, but will also instantly cause a seg-fault at that line of code, when executing.
I don't know what's really going on there, but auto fixes it.
I'm all for standard guidelines improving code readability, but not for when those guidelines are actively counter productive.
https://docs.oracle.com/javase/8/docs/technotes/guides/jni/spec/functions.html#NewStringUTF
For whatever reason, it "should" be jstring, but jstring causes a segfault.
Worthy of filing a separate bug, but this demonstrably fixes the segfault.

source/code/source/playfab/PlayFabAndroidHttpPlugin.cpp
pgilmorepf: Ditto 1'st reply

source/code/source/playfab/PlayFabAndroidHttpPlugin.cpp
pgilmorepf: Ditto 1'st reply

source/code/source/playfab/PlayFabAndroidHttpPlugin.cpp
pgilmorepf: A few thoughts:
Right now, requestContainer is always defined, and must always be defined, because we use it below as well.
Here, I'm making a very strong assumption about how I can convert requestContainer from BaseCallRequestContainer to CallRequestContainer, which is an assumption made in every single HTTP layer at present.
Exaggerating this assumption here doesn't fundamentally change anything at present, because that assumption is everywhere, in every HTTP layer, and if that assumption were false, it'd break in dozens of other places too.
This is already a known "we'll have to fix this later" task that's already filed.

source/build/Android/gradle/wrapper/gradle-wrapper.properties
pgilmorepf: I don't know enough about Android Studio to know for sure.
I feel the safest assumption today is to just assume the initial creator knew what they were doing.
...
According to a casual search, the first, highest-rated responses to multiple copies of this question seem to indicate that this file should be included, and not ignored.

source/code/source/playfab/PlayFabAndroidHttpPlugin.cpp
toddbellMSFT: requestContainer is a nullptr in the RequestTask object BUT 1.) it's only used in this plugin 2.) the only place where this object is used, we immediately call Initialize. Though note that this is in a try/catch block above that doesn't exit if it catches an exception. In some HIGHLY unlikely scenario where the construction worked but the Init didn't, the requestTask.requestContainer could still be a nullptr. 

What's that task number?

source/code/source/playfab/PlayFabAndroidHttpPlugin.cpp
toddbellMSFT: https://dev.azure.com/playfab/PlayFab/_workitems/edit/38975

pgilmorepf:
Discussed and responded to nullptr, which is covered by a future task.
Please confirm ready to land.

MichaelMackendMS:
>         JNIEnv* jniEnv = requestTask.impl->JniEvn();

typo

---
Refers to: source/code/source/playfab/PlayFabAndroidHttpPlugin.cpp:407 in de71e8b. [](commit_id = de71e8bc3f562b13ca292c67a1497ae1a35c88ee, deletion_comment = False)

PR #59, opened 2019-12-05
Enable support for Platform Specific login

pgilmorepf:
... with separate implementation per-platform

### DIFF
--- a/source/test/TestApp/PlayFabTestMultiUserInstance.cpp
+++ b/source/test/TestApp/PlayFabTestMultiUserInstance.cpp
@@ -14,12 +14,6 @@ using namespace ClientModels;
 
 namespace PlayFabUnit
 {
-    std::string multiUserInstanceTitleId;
-    void PlayFabTestMultiUserInstance::SetTitleInfo(const TestTitleData& testTitleData)
-    {
-        multiUserInstanceTitleId = testTitleData.titleId;
-    }
-
     /// <summary>
     /// CLIENT API
     /// Try to log in two users simultaneously using instance APIs.
@@ -100,7 +94,7 @@ namespace PlayFabUnit
     void PlayFabTestMultiUserInstance::ClassSetUp()
     {
         settings2 = std::make_shared<PlayFab::PlayFabApiSettings>();
-        settings2->titleId = multiUserInstanceTitleId;
+        settings2->titleId = testTitleData.titleId;
 
         // Create API handles for all users.
         multiUser1ClientApi = std::make_shared<PlayFabClientInstanceAPI>();
@@ -109,7 +103,7 @@ namespace PlayFabUnit
 
     void PlayFabTestMultiUserInstance::SetUp(TestContext& /*testContext*/)
     {
-        PlayFab::PlayFabSettings::staticSettings->titleId = multiUserInstanceTitleId;
+        PlayFab::PlayFabSettings::staticSettings->titleId = testTitleData.titleId;
 
         // Make sure PlayFab state is clean.
         PlayFabSettings::ForgetAllCredentials();
--- a/source/test/TestApp/PlayFabTestMultiUserInstance.cpp
+++ b/source/test/TestApp/PlayFabTestMultiUserInstance.cpp
@@ -14,6 +14,12 @@ using namespace ClientModels;
 
 namespace PlayFabUnit
 {
+    std::string multiUserInstanceTitleId;
+    void PlayFabTestMultiUserInstance::SetTitleInfo(TestTitleData& testTitleData)
+    {
+        multiUserInstanceTitleId = testTitleData.titleId;
+    }
+
     /// <summary>
     /// CLIENT API
     /// Try to log in two users simultaneously using instance APIs.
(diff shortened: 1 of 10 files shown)


source/test/TestApp/PlayFabTestMultiUserInstance.cpp
```diff
@@ -14,6 +14,12 @@ using namespace ClientModels;
 
 namespace PlayFabUnit
 {
+    std::string multiUserInstanceTitleId;
+    void PlayFabTestMultiUserInstance::SetTitleInfo(TestTitleData& testTitleData)
```
MichaelMackendMS: Should be const&

MichaelMackendMS:
>     class PlayFabTestMultiUserInstance : public TestCase

Let's rename this to PlayFabMultiUserInstanceLoginTest or something else that better identifies this as a test case. 

---
Refers to: source/test/TestApp/PlayFabTestMultiUserInstance.h:23 in ee65ad9. [](commit_id = ee65ad9485c9f11f473f4b2de88a67c1d8c1e76b, deletion_comment = False)

MichaelMackendMS:
>         std::shared_ptr<PlayFab::PlayFabApiSettings> settings2;

Where is settings1?

---
Refers to: source/test/TestApp/PlayFabTestMultiUserInstance.h:31 in ee65ad9. [](commit_id = ee65ad9485c9f11f473f4b2de88a67c1d8c1e76b, deletion_comment = False)

source/test/TestApp/PlayFabTestMultiUserInstance.h
```diff
@@ -47,6 +47,8 @@ namespace PlayFabUnit
         void AddTests() override;
 
     public:
+        void SetTitleInfo(TestTitleData& testTitleData);
```
MichaelMackendMS: const&

pgilmorepf:
>     class PlayFabTestMultiUserInstance : public TestCase

Out of scope for this PR. I didn't touch much here, except to add SetTitleInfo.

---
In reply to: [561922431](https://github.com/PlayFab/XPlatCoreTemplate/pull/59#issuecomment-561922431) [](ancestors = 561922431)

---
Refers to: source/test/TestApp/PlayFabTestMultiUserInstance.h:23 in ee65ad9. [](commit_id = ee65ad9485c9f11f473f4b2de88a67c1d8c1e76b, deletion_comment = False)

source/test/TestApp/PlayFabTestMultiUserInstance.cpp
```diff
@@ -14,6 +14,12 @@ using namespace ClientModels;
 
 namespace PlayFabUnit
 {
+    std::string multiUserInstanceTitleId;
```
MichaelMackendMS: Any specific reason we are storing this as a global here instead of a member variable?

pgilmorepf:
>         std::shared_ptr<PlayFab::PlayFabApiSettings> settings2;

Instance 1 uses the default settings.
Settings2 is for instance 2.
That's explicitly what's being tested, default settings vs explicit settings.

---
In reply to: [561922620](https://github.com/PlayFab/XPlatCoreTemplate/pull/59#issuecomment-561922620) [](ancestors = 561922620)

---
Refers to: source/test/TestApp/PlayFabTestMultiUserInstance.h:31 in ee65ad9. [](commit_id = ee65ad9485c9f11f473f4b2de88a67c1d8c1e76b, deletion_comment = False)

source/test/TestApp/PlayFabTestMultiUserInstance.cpp
pgilmorepf: It's not really global since it's only defined in the cpp.
I just wanted a temp variable to carry the titleId from SetTitleInfo to the SetUp function, which is called much later.

---
In reply to: [354069484](https://github.com/PlayFab/XPlatCoreTemplate/pull/59#discussion_r354069484) [](ancestors = 354069484)

pgilmorepf: Alright, I'll fix.

---
In reply to: [354069858](https://github.com/PlayFab/XPlatCoreTemplate/pull/59#discussion_r354069858) [](ancestors = 354069858,354069484)

source/test/TestApp/PlayFabTestMultiUserStatic.h
```diff
@@ -45,6 +45,8 @@ namespace PlayFabUnit
         void AddTests() override;
 
     public:
+        void SetTitleInfo(TestTitleData& testTitleData);
```
MichaelMackendMS: const&

source/test/TestApp/PlayFabTestMultiUserStatic.cpp
```diff
@@ -14,6 +14,12 @@ using namespace ClientModels;
 
 namespace PlayFabUnit
 {
+    std::string multiUserStaticTitleId;
```
MichaelMackendMS: ditto from MultiUserInstance

PR #60, opened 2019-12-06
Remove many shared_ptr<Type*>

pgilmorepf:
Because that's a massive memory leak
(Destroying the pointer to an obj doesn't destroy the obj)

**Related fixes:**
Fix all the places where we were relying on that memory leak (Mostly tests)
Do NOT make shared_ptr's to an object whose life-cycle we don't control (TestContext)

**Related refactoring:**
Track the N'th event a little more carefully.
Removing all static fields from PlayFabEventTest

**Good practice and cleanliness refactoring:**
Remove all trailing spaces in all files in the project
Converting all shared_ptr(new Thing()) to make_shared()
Standardize spacing (vs ctrl+k+d)

### DIFF
--- a/source/code/source/playfab/PlayFabEventPipeline.cpp
+++ b/source/code/source/playfab/PlayFabEventPipeline.cpp
@@ -155,7 +155,7 @@ namespace PlayFab
         using clock = std::chrono::steady_clock;
         using Result = PlayFabEventBuffer::EventConsumingResult;
         std::shared_ptr<const IPlayFabEmitEventRequest> request;
-        size_t batchCounter = 0; // used to track uniqueness of batches in the map
+        size_t batchCounter = 1; // used to track uniqueness of batches in the map
         std::chrono::steady_clock::time_point momentBatchStarted; // used to track when a currently assembled batch got its first event
 
         while (this->isWorkerThreadRunning)
--- a/source/code/source/playfab/QoS/PlayFabQoSApi.cpp
+++ b/source/code/source/playfab/QoS/PlayFabQoSApi.cpp
@@ -37,7 +37,7 @@ namespace PlayFab
                 {
                     PlayFabSettings::globalErrorHandler(container.errorWrapper, container.GetCustomData());
                 }
-                
+
                 if (container.errorCallback != nullptr)
                 {
                     container.errorCallback(container.errorWrapper, container.GetCustomData());
@@ -137,7 +137,7 @@ namespace PlayFab
         {
             Json::Value value;
             value["ErrorCode"] = Json::Value(result.errorCode);
-            
+
             Json::Value each_regionCenterResult;
             for (int i = 0; i < result.regionResults.size(); ++i)
             {
@@ -185,7 +185,7 @@ namespace PlayFab
             ListQosServersRequest request;
             PlayFabMultiplayerAPI::ListQosServers(request, ListQosServersSuccessCallBack, ListQosServersFailureCallBack, reinterpret_cast<void*>(this));
         }
-        
+
         void PlayFabQoSApi::ListQosServersSuccessCallBack(const ListQosServersResponse& result, void* customData)
         {
             // Custom data received is a pointer to our api object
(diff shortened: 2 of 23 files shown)


source/code/source/playfab/PlayFabEventPipeline.cpp
```diff
@@ -155,7 +155,7 @@ namespace PlayFab
         using clock = std::chrono::steady_clock;
         using Result = PlayFabEventBuffer::EventConsumingResult;
         std::shared_ptr<const IPlayFabEmitEventRequest> request;
-        size_t batchCounter = 0; // used to track uniqueness of batches in the map
+        size_t batchCounter = 1; // used to track uniqueness of batches in the map
```
toddbellMSFT: interesting to start from expecting a single batch to start. i think it does make sense, but it's a bit of a change to go from nothing (0 when we initialize) to something (well we will probably put something on there anyway).

source/test/TestApp/PlayFabEventTest.cpp
```diff
@@ -68,15 +68,15 @@ namespace PlayFabUnit
 
         // TODO: Bug 38165 automated builds either time out here or tests will never complete.
         // send several events
-        //for (int i = 0; i < 2; i++)
-        //{
+        for (int i = 0; i < 2; i++)
```
toddbellMSFT: YES!

source/code/source/playfab/QoS/PlayFabQoSApi.cpp
```diff
@@ -247,7 +247,7 @@ namespace PlayFab
             // Setup sockets based on the number of threads
             for (unsigned int i = 0; i < numThreads; ++i)
             {
-                shared_ptr<QoSSocket> socket = shared_ptr<QoSSocket>(new QoSSocket());
+                shared_ptr<QoSSocket> socket = make_shared<QoSSocket>();
```
Vicodex: >make_shared<QoSSocket>(); [](start = 47, length = 25)

Change if you want, but i think we were supporting Std-11 until now in Xplat SDK

source/code/source/playfab/PlayFabEventPipeline.cpp
pgilmorepf: The reason for this change was actually for a different reason:
This is cast to a void*, where the value of 0 implies "nullptr" which was confusing me as I was debugging.
There's no harm in it being 1, and it was easier for me to interpret what customData was while debugging.

---
In reply to: [355049528](https://github.com/PlayFab/XPlatCoreTemplate/pull/60#discussion_r355049528) [](ancestors = 355049528)

source/code/source/playfab/QoS/PlayFabQoSApi.cpp
pgilmorepf: This isn't the first make_shared in the published/central part of the SDK.
We've not been testing C++11, so many make_shared's have already found their way into the SDK.
We could probably restore C++11 if customers declare the need, but for now, the ship has already sailed.

---
In reply to: [355050637](https://github.com/PlayFab/XPlatCoreTemplate/pull/60#discussion_r355050637) [](ancestors = 355050637)

source/test/TestApp/PlayFabTestAlloc.cpp
```diff
@@ -67,10 +68,28 @@ namespace PlayFabUnit
...
+        std::shared_ptr<PlayFab::PlayFabAuthenticationInstanceAPI> testSpApi = std::make_shared<PlayFab::PlayFabAuthenticationInstanceAPI>();
+        testSpApi.reset();
+
+        std::shared_ptr<PlayFab::PlayFabAuthenticationInstanceAPI> testNSpApi = std::shared_ptr<PlayFab::PlayFabAuthenticationInstanceAPI>(new PlayFab::PlayFabAuthenticationInstanceAPI());
+        testNSpApi.reset();
+
+        testContext.Pass();
+    }
+
     void PlayFabTestAlloc::AddTests()
```
MichaelMackendMS: This is a strange test.  What is the motivation behind testing allocation w/wo shared pointer etc?  This seems more like a test of the functionality of shared_ptr than it is a test of allocating a PlayFabAuthenticationInstanceIApi...

MichaelMackendMS:
Good PR.  This had some terrifying misuse of shared_ptr.  :o

source/test/TestApp/PlayFabTestAlloc.cpp
pgilmorepf: The motivation is that the objects in this file have been known to seg-fault when being destroyed.
This was the big 3 hour full-team discussion about how the exe and the lib think the object is a different size.

PR #62, opened 2019-12-10
Event API Testing and fixing

pgilmorepf:
Address platform specific thread limits
Fix some comments that are out of date
EventTests perform their own login, and better isolate test settings
These tests work on the Switch now, so removed that platform filter

### DIFF
--- a/source/code/include/playfab/PlayFabEvent.h
+++ b/source/code/include/playfab/PlayFabEvent.h
@@ -16,8 +16,8 @@ namespace PlayFab
     enum class PlayFabEventType
     {
         Default, // Default type (e.g. the one set by global configuration)
-        Lightweight, // Event is meant to bypass processing on the PlayFab server side and be sent instead directly to One Collector (1DS).
-        Heavyweight // Event is meant to be sent to the PlayFab server.
+        Lightweight, // WriteTelemetryEvents
+        Heavyweight // WriteEvents
     };
 
     /// <summary>
--- a/source/code/include/playfab/PlayFabEvent.h
+++ b/source/code/include/playfab/PlayFabEvent.h
@@ -16,8 +16,8 @@ namespace PlayFab
     enum class PlayFabEventType
     {
         Default, // Default type (e.g. the one set by global configuration)
-        Lightweight, // WriteTelemetryEvents
-        Heavyweight // WriteEvents
+        Lightweight, // Event will be sent using WriteTelemetryEvents method in EventsAPI
+        Heavyweight // Event will be sent using WriteEvents method in EventsAPI
     };
 
     /// <summary>
--- a/source/test/TestApp/PlayFabEventTest.cpp
+++ b/source/test/TestApp/PlayFabEventTest.cpp
@@ -59,7 +59,7 @@ namespace PlayFabUnit
             ApiCallback(&PlayFabEventTest::OnErrorSharedCallback),
             &testContext);
     }
-    void PlayFabEventTest::OnLogin(const PlayFab::ClientModels::LoginResult& result, void* customData)
+    void PlayFabEventTest::OnLogin(const PlayFab::ClientModels::LoginResult& /*result*/, void* customData)
     {
         TestContext* testContext = static_cast<TestContext*>(customData);
         testContext->Pass();
@@ -322,7 +322,7 @@ namespace PlayFabUnit
         PlayFabSettings::ForgetAllCredentials();
     }
 
-    void PlayFabEventTest::SetUp(TestContext& testContext)
+    void PlayFabEventTest::SetUp(TestContext& /*testContext*/)
     {
         PlayFabSettings::staticSettings->titleId = testTitleData.titleId;
         // Reset event test values.
--- a/source/test/TestApp/PlayFabEventTest.cpp
+++ b/source/test/TestApp/PlayFabEventTest.cpp
@@ -301,11 +301,13 @@ namespace PlayFabUnit
 
     void PlayFabEventTest::AddTests()
     {
+        AddTest("BasicLogin", &PlayFabEventTest::BasicLogin);
+
         // TODO: Fix whatever limitation causes this test to fail for these platforms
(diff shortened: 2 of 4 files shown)


source/code/include/playfab/PlayFabEvent.h
```diff
@@ -16,8 +16,8 @@ namespace PlayFab
     enum class PlayFabEventType
     {
         Default, // Default type (e.g. the one set by global configuration)
-        Lightweight, // Event is meant to bypass processing on the PlayFab server side and be sent instead directly to One Collector (1DS).
-        Heavyweight // Event is meant to be sent to the PlayFab server.
+        Lightweight, // WriteTelemetryEvents
```
Vicodex: >WriteTelemetryEvents [](start = 24, length = 20)

I agree that for someone like us it's pretty obvious what these comments mean :) For someone else though, who just picked up our SDK it might be more useful if we add a few more words here. For example, "Event will be sent using WriteTelemetryEvents method in EventsAPI"

source/test/TestApp/PlayFabEventTest.cpp
```diff
@@ -262,6 +284,7 @@ namespace PlayFabUnit
 
     void PlayFabEventTest::ManyThreadsLowEventsPerTest(TestContext& testContext)
     {
+        // NOTE: Some platforms have hard limits on the maximum number of threads (This number found to be experimentally safe)
```
Vicodex: >T [](start = 83, length = 1)

this "T" may be smaller :)

source/test/TestApp/PlayFabEventTest.cpp
```diff
@@ -43,6 +43,28 @@ namespace PlayFabUnit
     constexpr int CLOUDSCRIPT_TIMEOUT_MS = 10000;
     constexpr int CLOUDSCRIPT_TIMEOUT_INCREMENT = 100;
 
+    void PlayFabEventTest::OnErrorSharedCallback(const PlayFab::PlayFabError& error, void* customData)
+    {
+        TestContext* testContext = reinterpret_cast<TestContext*>(customData);
```
MichaelMackendMS: nit: prefer static_cast

source/test/TestApp/PlayFabEventTest.cpp
```diff
@@ -222,19 +244,19 @@ namespace PlayFabUnit
...
                 {
-                    std::shared_ptr<PlayFabEventAPI> api = SetupEventTest();
                     for (uint32_t i = 0; i < pNumEventsPerThread; ++i)
                     {
                         api->EmitEvent(MakeEvent(PlayFabEventType::Default),
                             [&eventsRemaining, this]
                         (std::shared_ptr<const PlayFab::IPlayFabEvent>, std::shared_ptr<const PlayFab::IPlayFabEmitEventResponse>)
                         {
-                            if (--eventsRemaining == 0)
+                            if (--eventsRemaining <= 0)
```
MichaelMackendMS: this is alarming.  why would we get to negative events?

pgilmorepf: So, I was trying to be proactive, and also guessing at possible causes for failure in the PlayStation.
It didn't pan out. This test passes on everything but PS, where it just crashes.
BUT, I think this is still a good because:
1. If Pass() gets called twice, it will result in a test failure (cant call pass twice)
2. If some future change or circumstance causes more than the expected events to occur, the test will start failing (can't call Pass() twice).
3. Without this change, if more than the expected number of events occur, they'll be ignored and the test will pass, because we only call Pass() when ==0, which will only be once.

---
In reply to: [356226877](https://github.com/PlayFab/XPlatCoreTemplate/pull/62#discussion_r356226877) [](ancestors = 356226877)

source/test/TestApp/PlayFabEventTest.cpp
pgilmorepf: I'm going to by design this one.
It's the test framework, not the SDK, so performance isn't nearly as big of a factor (not zero, but...).
And, tests need to be a little more lenient/flexible by their nature of "anticipating incorrect code and or failures".

---
In reply to: [356223570](https://github.com/PlayFab/XPlatCoreTemplate/pull/62#discussion_r356223570) [](ancestors = 356223570)

source/code/include/playfab/PlayFabEvent.h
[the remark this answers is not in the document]
pgilmorepf: I think you were looking at an older version. I updated the comments per Victor's suggestion.

---
In reply to: [356211782](https://github.com/PlayFab/XPlatCoreTemplate/pull/62#discussion_r356211782) [](ancestors = 356211782)

source/test/TestApp/PlayFabEventTest.cpp
MichaelMackendMS: Wouldn't it be a red flag if more than the expected number of events occur?

source/test/TestApp/PlayFabEventTest.cpp
MichaelMackendMS: lol. Youu actually used static_cast in this same PR on line 64.

pgilmorepf: After in-person discussion, I think I'll just file a separate task to re-evaluate every reinterpret_cast and try and convert them all to static_cast.

---
In reply to: [356244857](https://github.com/PlayFab/XPlatCoreTemplate/pull/62#discussion_r356244857) [](ancestors = 356244857,356223570)

source/test/TestApp/PlayFabEventTest.cpp
pgilmorepf: Yes, that's exactly right. And now, and most importantly unlike before; the test will fail if that occurs.

source/test/TestApp/PlayFabEventTest.cpp
MichaelMackendMS: hmm

source/test/TestApp/PlayFabEventTest.cpp
MichaelMackendMS: ah ok

PR #63, opened 2019-12-11
Update PlayFabAndroidHttpPlugin.cpp

toddbellMSFT:
We never added a requestId to the requestTask on Android before we send out the request. So when we get the request back we ended up with an empty id. 

This fix will follow Window's paradigm and add the requestId as a parameter to ProcessResponse. We attempt to ask the jni to process the httpHeader (and explicitly get the X-RequestId header) and pass that back.

### DIFF
--- a/source/code/source/playfab/PlayFabAndroidHttpPlugin.cpp
+++ b/source/code/source/playfab/PlayFabAndroidHttpPlugin.cpp
@@ -575,7 +575,7 @@ namespace PlayFab
                             std::vector<uint8_t> requestIdBuffer(static_cast<size_t>(requestIdSize));
                             jniEnv->GetByteArrayRegion(requestIdBytes, 0, requestIdSize, reinterpret_cast<jbyte*>(requestIdBuffer.data()));
                             std::string reqId(reinterpret_cast<const char*>(requestIdBuffer.data()), requestIdBuffer.size());
-                            requestContainer.SetRequestID(reqId)
+                            requestContainer.SetRequestID(reqId);
                         }
                     }
                 }
--- a/source/code/source/playfab/PlayFabAndroidHttpPlugin.cpp
+++ b/source/code/source/playfab/PlayFabAndroidHttpPlugin.cpp
@@ -251,6 +251,11 @@ namespace PlayFab
         if (container != nullptr && container->HandleInvalidSettings())
         {
             std::shared_ptr<RequestTask> requestTask = nullptr;
+            { // LOCK httpRequestMutex
+                std::unique_lock<std::mutex> lock(httpRequestMutex);
+                container->requestId = pendingRequests.size()+1;
+            } // UNLOCK httpRequestMutex
+
             try
             {
                 requestTask = std::make_shared<RequestTask>();
@@ -388,6 +393,7 @@ namespace PlayFab
                     else
                     { // LOCK httpRequestMutex
                         std::unique_lock<std::mutex> lock(httpRequestMutex);
+                        this->requestingTask->impl = nullptr;
                         this->pendingResults.emplace_back(std::move(this->requestingTask));
                     } // UNLOCK httpRequestMutex
 
--- a/source/code/source/playfab/PlayFabAndroidHttpPlugin.cpp
+++ b/source/code/source/playfab/PlayFabAndroidHttpPlugin.cpp
@@ -562,7 +562,6 @@ namespace PlayFab
                 }
             }
 
-            std::string requestId = "initializingRequestId";
             {
                 methodId = jniEnv->GetMethodID(GetHelper().GetHttpRequestClass(), "getRequestIdFromHeader","()[B");
                 if (methodId)
@@ -576,7 +575,7 @@ namespace PlayFab
                             std::vector<uint8_t> requestIdBuffer(static_cast<size_t>(requestIdSize));
                             jniEnv->GetByteArrayRegion(requestIdBytes, 0, requestIdSize, reinterpret_cast<jbyte*>(requestIdBuffer.data()));
                             std::string reqId(reinterpret_cast<const char*>(requestIdBuffer.data()), requestIdBuffer.size());
-                            requestId = reqId;
+                            requestContainer.SetRequestID(reqId)
                         }
                     }
                 }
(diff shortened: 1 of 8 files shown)


source/code/source/playfab/PlayFabAndroidHttpPlugin.cpp
```diff
@@ -251,6 +251,11 @@ namespace PlayFab
         if (container != nullptr && container->HandleInvalidSettings())
         {
             std::shared_ptr<RequestTask> requestTask = nullptr;
+            { // LOCK httpRequestMutex
+                std::unique_lock<std::mutex> lock(httpRequestMutex);
+                container->requestId = pendingRequests.size()+1;
```
pgilmorepf: This is not the correct requestId.
The RequestId is extracted from the http response headers from the PlayFab API call.
It's information that the customer sends us when they experience a bug that they can't fix themselves, so we can look up their call in our internal logs.
It's not just an arbitrary value.

source/code/source/playfab/PlayFabAndroidHttpPlugin.cpp
```diff
@@ -562,6 +562,26 @@ namespace PlayFab
...
+                    jbyteArray requestIdBytes = (jbyteArray)jniEnv->CallObjectMethod(httpRequestObject, methodId);
+                    if(requestIdBytes != nullptr)
+                    {
+                        int requestIdSize = jniEnv->GetArrayLength(requestIdBytes);
+                        if(requestIdSize > 0)
+                        {
+                            std::vector<uint8_t> requestIdBuffer(static_cast<size_t>(requestIdSize));
+                            jniEnv->GetByteArrayRegion(requestIdBytes, 0, requestIdSize, reinterpret_cast<jbyte*>(requestIdBuffer.data()));
+                            std::string reqId(reinterpret_cast<const char*>(requestIdBuffer.data()), requestIdBuffer.size());
+                            requestId = reqId;
```
pgilmorepf: Delete requestId temp variable, and change this line to:
requestContainer.SetRequestID(reqId)

source/code/source/playfab/PlayFabAndroidHttpPlugin.cpp
```diff
@@ -578,7 +598,7 @@ namespace PlayFab
 
                             std::string body(reinterpret_cast<const char* >(bodyBuffer.data()), bodyBuffer.size());
                             requestContainer.responseString = body;
-                            ProcessResponse(*(this->requestingTask), static_cast<const int>(httpCode));
+                            ProcessResponse(*(this->requestingTask), static_cast<const int>(httpCode), requestId);
```
pgilmorepf: requestContainer.SetRequestID(reqId)
makes this line change un-necessary

source/code/source/playfab/PlayFabAndroidHttpPlugin.cpp
```diff
@@ -598,7 +618,8 @@ namespace PlayFab
     void PlayFabAndroidHttpPlugin::SetResponseAsBadRequest(RequestTask& requestTask)
     {
         CallRequestContainer& requestContainer = this->requestingTask->RequestContainer();
-        ProcessResponse(*(this->requestingTask), 400); // 400 Bad Request
+        std::string temp = "NoID-BadRequest";
+        ProcessResponse(*(this->requestingTask), 400, temp); // 400 Bad Request
```
pgilmorepf: requestContainer.SetRequestID(reqId)
makes this line change un-necessary

source/code/source/playfab/PlayFabAndroidHttpPlugin.cpp
```diff
@@ -598,7 +618,8 @@ namespace PlayFab
     void PlayFabAndroidHttpPlugin::SetResponseAsBadRequest(RequestTask& requestTask)
     {
         CallRequestContainer& requestContainer = this->requestingTask->RequestContainer();
-        ProcessResponse(*(this->requestingTask), 400); // 400 Bad Request
+        std::string temp = "NoID-BadRequest";
```
pgilmorepf: This "id not set" default be set via requestContainer.SetRequestID() if there's a failure above.

pgilmorepf: Though, it's not in-line with the "" that every other platform uses...

source/code/source/playfab/PlayFabAndroidHttpPlugin.cpp
```diff
@@ -653,15 +674,15 @@ namespace PlayFab
         return false;
     }
 
-    void PlayFabAndroidHttpPlugin::ProcessResponse(RequestTask& requestTask, const int httpCode)
+    void PlayFabAndroidHttpPlugin::ProcessResponse(RequestTask& requestTask, const int httpCode, const std::string& requestId)
```
pgilmorepf: requestContainer.SetRequestID(reqId)
makes this line change un-necessary

source/code/source/playfab/PlayFabAndroidHttpPlugin.cpp
```diff
@@ -653,15 +674,15 @@ namespace PlayFab
...
     {
         CallRequestContainer& requestContainer = requestTask.RequestContainer();
         Json::CharReaderBuilder jsonReaderFactory;
         std::unique_ptr<Json::CharReader> jsonReader(jsonReaderFactory.newCharReader());
         JSONCPP_STRING jsonParseErrors;
         const bool parsedSuccessfully = jsonReader->parse(requestContainer.responseString.c_str(), requestContainer.responseString.c_str() + requestContainer.responseString.length(), &requestContainer.responseJson, &jsonParseErrors);
 
-        requestContainer.errorWrapper.RequestId = requestContainer.GetRequestId();
+        requestContainer.errorWrapper.RequestId = requestId;
```
pgilmorepf: requestContainer.SetRequestID(reqId)
makes this line change un-necessary

toddbellMSFT: Surprisingly, this is still a necessary change. The requestContainer is not the errorWrapper, without this line, the PlayFabError does not get the RequestId. 

---
In reply to: [357435817](https://github.com/PlayFab/XPlatCoreTemplate/pull/63#discussion_r357435817) [](ancestors = 357435817)

toddbellMSFT: Updated SetRequestId to also set the container. NOW this line is no longer required (which also looked so weird to have an object set something to itself).

---
In reply to: [357439799](https://github.com/PlayFab/XPlatCoreTemplate/pull/63#discussion_r357439799) [](ancestors = 357439799,357435817)

toddbellMSFT:
This branch was merged, but managed to stay around as an open PR that was 2 commits behind master.

PR #65, opened 2019-12-13
Time overhaul

pgilmorepf:
Some platforms have feature limitations when it comes to time, so we separate into distinct needs:

- The ability to accurately measure a duration
- The ability to serialize UTC timestamps to/from PlayFab
- Multi-Thread compatible conversion funcs in all time related code
- Consistency between PlayFab server time and local device time (as UTC timezone), with a properly configured device

Fixed (often unrelated) some reinterpret_casts to static_casts when applicable
Updated the utilization of the feature-appropriate-timestamp-methods where needed (mostly test framework)
GetServerTime now outputs the local interpretation of the server time (for additional human verification)
GetServerTime now tests a range of "suspicious" outputs with the deserialized Server Time (basically check that it's not 1970)
And good descriptive and highly accurate names for all the methods.
A full set of conversion functions for useful utility.

### DIFF
--- a/source/code/include/playfab/PlayFabPlatformUtils.h
+++ b/source/code/include/playfab/PlayFabPlatformUtils.h
@@ -91,7 +91,7 @@ namespace PlayFab
         static_assert("You must request the Nintendo specific XPlat SDK from PlayFab support.");
 #else
         auto msClock = std::chrono::time_point_cast<std::chrono::milliseconds>(Clock::now());
-        return static_cast<Int64>(msClock.time_since_epoch().count());
+        return msClock.time_since_epoch().count();
 #endif
     }
 
--- a/source/code/include/playfab/PlayFabPlatformUtils.h
+++ b/source/code/include/playfab/PlayFabPlatformUtils.h
@@ -39,7 +39,7 @@ namespace PlayFab
 
     inline tm TimeTToUtcTm(time_t input)
     {
-        tm timeInfo;
+        tm timeInfo{ 0 };
 #if defined(PLAYFAB_PLATFORM_PLAYSTATION)
         gmtime_s(&input, &timeInfo);
 #elif defined(PLAYFAB_PLATFORM_WINDOWS) || defined(PLAYFAB_PLATFORM_XBOX)
--- a/source/code/include/playfab/PlayFabPlatformUtils.h
+++ b/source/code/include/playfab/PlayFabPlatformUtils.h
@@ -104,7 +104,7 @@ namespace PlayFab
 
     inline tm Iso8601StringToTm(const std::string& utcString)
     {
-        tm timeInfo;
+        tm timeInfo{ 0 };
         std::istringstream iss(utcString);
         iss >> std::get_time(&timeInfo, TIMESTAMP_READ_FORMAT);
         return timeInfo;
--- a/source/code/include/playfab/PlayFabPlatformUtils.h
+++ b/source/code/include/playfab/PlayFabPlatformUtils.h
@@ -17,6 +17,7 @@ namespace PlayFab
     // The primary purpose of these format strings is to communicate to and from the PlayFab server with consistent accuracy across platforms supported by this SDK
     constexpr char TIMESTAMP_READ_FORMAT[] = "%Y-%m-%dT%T";
     constexpr char TIMESTAMP_WRITE_FORMAT[] = "%Y-%m-%dT%H:%M:%S.000Z";
+    constexpr int TIMESTAMP_BUFFER_SIZE = 64; // Arbitrary number sufficiently large enough to contain the timestamp strings sent by PlayFab server
 
     // Initialize may be required on some platforms
     inline void InitializeClock()
@@ -97,8 +98,8 @@ namespace PlayFab
     // Time Serialization
     inline std::string UtcTmToIso8601String(const tm& input)
     {
-        char buff[64];
-        strftime(buff, 64, TIMESTAMP_WRITE_FORMAT, &input);
+        char buff[TIMESTAMP_BUFFER_SIZE];
+        strftime(buff, TIMESTAMP_BUFFER_SIZE, TIMESTAMP_WRITE_FORMAT, &input);
         return buff;
     }
 
--- a/source/code/include/playfab/PlayFabPlatformUtils.h
+++ b/source/code/include/playfab/PlayFabPlatformUtils.h
@@ -40,7 +40,7 @@ namespace PlayFab
     inline tm TimeTToUtcTm(time_t input)
     {
         tm timeInfo;
-#if defined(PLAYFAB_PLATFORM_WINDOWS) || defined(PLAYFAB_PLATFORM_XBOX)
+#if defined(PLAYFAB_PLATFORM_WINDOWS) || defined(PLAYFAB_PLATFORM_XBOX) || defined(PLAYFAB_PLATFORM_PLAYSTATION)
 #define gmtime_r gmtime_s
 #endif
         gmtime_r(&input, &timeInfo);
@@ -49,10 +49,14 @@ namespace PlayFab
 
     inline time_t UtcTmToTimeT(tm input)
     {
+#if defined(PLAYFAB_PLATFORM_PLAYSTATION)
+        return mktime(&input);
+#else
 #if defined(PLAYFAB_PLATFORM_WINDOWS) || defined(PLAYFAB_PLATFORM_XBOX)
 #define timegm _mkgmtime
 #endif
         return timegm(&input);
+#endif
     }
 
     inline tm TimePointToUtcTm(TimePoint input)
@@ -84,7 +88,7 @@ namespace PlayFab
         static_assert("You must request the Nintendo specific XPlat SDK from PlayFab support.");
 #else
         auto msClock = std::chrono::time_point_cast<std::chrono::milliseconds>(Clock::now());
-        return msClock.time_since_epoch().count();
+        return static_cast<long>(msClock.time_since_epoch().count());
 #endif
     }
 
--- a/source/code/include/playfab/PlayFabPlatformUtils.h
+++ b/source/code/include/playfab/PlayFabPlatformUtils.h
@@ -85,13 +85,13 @@ namespace PlayFab
     }
 
     // Get a tick count that represents now in milliseconds (not useful for absolute time)
-    inline long GetMilliTicks()
+    inline Int64 GetMilliTicks()
     {
 #if defined(PLAYFAB_PLATFORM_SWITCH)
         static_assert("You must request the Nintendo specific XPlat SDK from PlayFab support.");
 #else
         auto msClock = std::chrono::time_point_cast<std::chrono::milliseconds>(Clock::now());
-        return static_cast<long>(msClock.time_since_epoch().count());
+        return static_cast<Int64>(msClock.time_since_epoch().count());
 #endif
     }
 
--- a/source/code/include/playfab/PlayFabPlatformUtils.h
+++ b/source/code/include/playfab/PlayFabPlatformUtils.h
@@ -7,53 +7,121 @@
 
 namespace PlayFab
 {
-    typedef std::chrono::time_point<std::chrono::system_clock> TimePoint;
+#if defined(PLAYFAB_PLATFORM_SWITCH)
+    static_assert("You must request the Nintendo specific XPlat SDK from PlayFab support.");
+#else
+    typedef std::chrono::system_clock Clock;
+    typedef std::chrono::time_point<Clock> TimePoint;
+#endif
 
-    inline TimePoint GetTimePointNow()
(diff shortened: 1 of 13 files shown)


pgilmorepf:
Confirmed tested on Switch.

pgilmorepf:
Confirmed tested on Windows.

pgilmorepf:
Confirmed tested on PlayStation.

pgilmorepf:
Confirm tested on Windows.

source/code/include/playfab/PlayFabPlatformUtils.h
```diff
@@ -7,53 +7,125 @@
...
-        return time(0);
+        return Clock::from_time_t(input);
     }
 
-    inline std::string LocalTimeTToUtcString(time_t now)
+    inline tm TimeTToUtcTm(time_t input)
     {
         tm timeInfo;
+#if defined(PLAYFAB_PLATFORM_WINDOWS) || defined(PLAYFAB_PLATFORM_XBOX) || defined(PLAYFAB_PLATFORM_PLAYSTATION)
+#define gmtime_r gmtime_s
```
ScottMunroMS: this concerns me. I'm pretty sure the windows UCRT version of gmtime_s has the order of parameters reversed (struct tm first then time_t) https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/gmtime-s-gmtime32-s-gmtime64-s?view=vs-2019

the return type appears to be different as well. How was this validated for windows?

in general I don't think the using preprocessor macros to override functions is a good idea here but it's got me extra confused with the flipped parameters

source/code/include/playfab/PlayFabPlatformUtils.h
```diff
@@ -7,53 +7,125 @@
...
+        static_assert("You must request the Nintendo specific XPlat SDK from PlayFab support.");
 #else
-        output = _mkgmtime(&timeStruct);
+        auto msClock = std::chrono::time_point_cast<std::chrono::milliseconds>(Clock::now());
+        return static_cast<long>(msClock.time_since_epoch().count());
 #endif
+    }
+
+    // Time Serialization
+    inline std::string UtcTmToIso8601String(tm input)
```
ScottMunroMS: I think it's idiomatic to avoid taking `tm` by value since it's a decent sized struct. recommend using a const-ref

source/code/include/playfab/PlayFabPlatformUtils.h
```diff
@@ -7,53 +7,125 @@
...
+        iss >> std::get_time(&timeInfo, TIMESTAMP_READ_FORMAT);
+        return timeInfo;
+    }
+
+    inline std::string TimeTToIso8601String(time_t input)
+    {
+        return UtcTmToIso8601String(TimeTToUtcTm(input));
+    }
+
+    inline time_t Iso8601StringToTimeT(std::string input)
```
ScottMunroMS: const-ref std::string

source/code/include/playfab/PlayFabPlatformUtils.h
```diff
@@ -7,53 +7,125 @@
...
+    }
+
+    // TODO: Invert this conversion at some point, and serialize the milliseconds as well
+    inline std::string TimePointToIso8601String(TimePoint input)
+    {
+        return UtcTmToIso8601String(TimePointToUtcTm(input));
+    }
 
-        return output;
+    inline TimePoint Iso8601StringToTimePoint(std::string input)
```
ScottMunroMS: const-ref std::string

source/code/include/playfab/PlayFabPlatformUtils.h
pgilmorepf: Tested on Windows vs 2017, and while I noticed the order was reversed, StackOverflow claimed it would work, and I verified that it did. I even confirmed that the struct was assigned properly, via automated tests, and manual debugging.
I was caught up in the beauty of it, but you're probably right, it's probably not safe overall.
I'll do a more normal version.

---
In reply to: [357515763](https://github.com/PlayFab/XPlatCoreTemplate/pull/65#discussion_r357515763) [](ancestors = 357515763)

pgilmorepf: Oh, it's worse than that... The order you see here is correct for PlayStation, and wrong for Windows. I made a mistake when I was testing on Windows (I swapped it locally, but didn't copy it back to the template).
I've got a new version, but sadly it won't be as pretty as this. But it'll work on everything.

---
In reply to: [357774542](https://github.com/PlayFab/XPlatCoreTemplate/pull/65#discussion_r357774542) [](ancestors = 357774542,357515763)

pgilmorepf:
After recent updates, confirmed tested on Windows, Switch, and PlayStation

source/code/include/playfab/PlayFabPlatformUtils.h
pgilmorepf: Scott, I'm marking this one resolved as I've addressed your original concern, re-tested carefully on multiple platforms, and confirmed that exactly this code works on all platforms.

---
In reply to: [357778543](https://github.com/PlayFab/XPlatCoreTemplate/pull/65#discussion_r357778543) [](ancestors = 357778543,357774542,357515763)

pgilmorepf: Ok, for reals this time... :D

---
In reply to: [357785771](https://github.com/PlayFab/XPlatCoreTemplate/pull/65#discussion_r357785771) [](ancestors = 357785771,357778543,357774542,357515763)

source/code/include/playfab/PlayFabPlatformUtils.h
```diff
@@ -7,53 +7,127 @@
...
-        output = mktime(&timeStruct);
-#elif defined(PLAYFAB_PLATFORM_IOS) || defined(PLAYFAB_PLATFORM_ANDROID) || defined(PLAYFAB_PLATFORM_LINUX)
-        gmtime_r(&output, &timeStruct);
+    inline time_t GetTimeTNow()
+    {
+        return TimePointToTimeT(GetTimePointNow());
+    }
+
+    // Get a tick count that represents now in milliseconds (not useful for absolute time)
+    inline long GetMilliTicks()
```
ScottMunroMS: recommend using uint64_t instead of long
1. fixed-width int type to make it clear what bounds the return value has (it's platform specific how big a long is)
2. the count of ticks should never be negative so returning a signed type doesn't make sense
3. if long is a 32-bit sized value it will overflow at least every 2 months (ish). This may not seem like a big deal but considering games can be suspended for LONG periods of time and resumed this is probably just better not left to chance

source/code/include/playfab/PlayFabPlatformUtils.h
```diff
@@ -7,53 +7,127 @@
...
-        output = _mkgmtime(&timeStruct);
+        auto msClock = std::chrono::time_point_cast<std::chrono::milliseconds>(Clock::now());
+        return static_cast<long>(msClock.time_since_epoch().count());
 #endif
+    }
+
+    // Time Serialization
+    inline std::string UtcTmToIso8601String(const tm& input)
+    {
+        char buff[64];
```
ScottMunroMS: magic numbers should be avoided. is there anyway you can tie this back to the WRITE_FORMAT?
e.g. char buff[] = "<an example timestring which is the max size that TIEMSTAMP_WRITE_FORMAT could write>";

source/code/include/playfab/PlayFabPlatformUtils.h
pgilmorepf: If I were going to use an alternate type, then I'd probably switch to UInt64, just to be consistent with the rest of the SDK (though I know you dislike that type).
I used a long, because long is the literal exact type that matches the underlying type for all the Clock implementations (though, you have to F12 into the std headers to find it).
Only one platform needed the static_cast<long> below, and that's only because it had typedef __rep long, rather than using long directly.
So, I understand your point, but if I choose to match the underlying implementation of the clocks, I should go with long, and if I want to consistently match the established types in PlayFab, I should use the UInt64 typedef.

---
In reply to: [357800254](https://github.com/PlayFab/XPlatCoreTemplate/pull/65#discussion_r357800254) [](ancestors = 357800254)

source/code/include/playfab/PlayFabPlatformUtils.h
pgilmorepf: 64 was chosen as "the first bigger 2^n number after the documented max number (~36 maybe)" for one of the methods used in the previous implementation (though I don't know which off-hand).
It's also similarly well-larger than the biggest string PlayFab is going to send.
It's always going to be fairly arbitrary, because the C# impl in PlayFab is not strictly sized like C++ is, so we just want to pick a number that's sufficiently flexible for what C# might do.
I do agree it could be a constant though.

---
In reply to: [357802305](https://github.com/PlayFab/XPlatCoreTemplate/pull/65#discussion_r357802305) [](ancestors = 357802305)

source/code/include/playfab/PlayFabPlatformUtils.h
ScottMunroMS: then I would prefer Uint64 or Int64
from looking at docs, std::chrono::milliseconds (which is the duration time you are using) says that it's underlying representation is a signed integer type of at least 45 bits. I think using Int64 or Uint64 is just a better way to communicate that [https://en.cppreference.com/w/cpp/chrono/duration](https://en.cppreference.com/w/cpp/chrono/duration)

---
In reply to: [357822540](https://github.com/PlayFab/XPlatCoreTemplate/pull/65#discussion_r357822540) [](ancestors = 357822540,357800254)

source/code/include/playfab/PlayFabPlatformUtils.h
```diff
@@ -85,13 +85,13 @@ namespace PlayFab
...
     // Get a tick count that represents now in milliseconds (not useful for absolute time)
-    inline long GetMilliTicks()
+    inline Int64 GetMilliTicks()
     {
 #if defined(PLAYFAB_PLATFORM_SWITCH)
         static_assert("You must request the Nintendo specific XPlat SDK from PlayFab support.");
 #else
         auto msClock = std::chrono::time_point_cast<std::chrono::milliseconds>(Clock::now());
-        return static_cast<long>(msClock.time_since_epoch().count());
+        return static_cast<Int64>(msClock.time_since_epoch().count());
```
ScottMunroMS: do you get some kind of compiler warning if you drop the static_cast? I would expect a long to implicitly cast to int64 just fine

ScottMunroMS: getting rid of casts is helpful because casts mask helpful warning messages (as discussed out of band)

pgilmorepf: When it was "long long" to "long" yes, but now that it's Int64, no.
Thanks!

---
In reply to: [357854735](https://github.com/PlayFab/XPlatCoreTemplate/pull/65#discussion_r357854735) [](ancestors = 357854735)

source/test/TestApp/PlayFabApiTest.cpp
```diff
@@ -276,6 +276,24 @@ namespace PlayFabUnit
     // Parameter types tested: DateTime
     void PlayFabApiTest::GetServerTime(TestContext& testContext)
     {
+        TimePoint nowTP = GetTimePointNow();
+        tm timeInfoTP = TimePointToUtcTm(nowTP);
+        int actualYearTP = timeInfoTP.tm_year + 1900; // Definition of tm_year is "years since 1900"
+        if (!(2019 <= actualYearTP && actualYearTP <= 2025))
```
Toddsa: NIT: Why assume this code wont be running in 5 years :)

pgilmorepf:
Build passes on Mac (but my local settings have problems executing)
Linux looks completely horribly busted, beyond the scope of this PR, and not remotely the fault of this PR.

source/test/TestApp/TestReport.h
```diff
@@ -57,11 +57,11 @@ namespace PlayFabUnit
     public:
         TestSuiteReport internalReport;
 
-        TestReport(std::string testName);
+        TestReport(const std::string& testName);
 
         void TestStarted();
 
-        void TestComplete(std::string testName, TestFinishState testFinishState, std::chrono::milliseconds testDurationMs, std::string message);
```
MichaelMackendMS: const& here too

source/test/TestApp/TestRunner.cpp
```diff
@@ -105,8 +105,8 @@ namespace PlayFabUnit
     {
         std::stringstream summaryStream;
 
-        PlayFab::TimePoint timeNow = PlayFab::GetTimePointNow();
-        PlayFab::TimePoint testStartTime, testEndTime;
+        Int64 timeNow = PlayFab::GetMilliTicks();
+        Int64 testStartTime, testEndTime;
```
MichaelMackendMS: It seems to me that testStartTime and testEndTime are only ever written to and never read.  Is there a logic problem here?

pgilmorepf:
Tasks filed for your feedback Michael, thanks.

PR #67, opened 2019-12-17
Minor fixes and more work on EventTests

pgilmorepf:
We're back to flagging EventTests as totally disabled
While they are working great in debug for a high % of runs for all platforms, they're failing 100% in release mode.
This set of test changes addresses some of the issues I found in this pass of test debugging, mostly around crashes if the test duration exceeds maximum (which seems to be a release-mode issue...?).
It also addresses some trivial CR comments that came in after previous commits were landed.

Note, some added commented code, which is needed for further investigation of EventTests.
It's my opinion at this time, based on lots of analysis and debugging, that the event pipeline is in good shape, but the tests still need an enormous amount of work.

### DIFF
--- a/source/test/TestApp/PlayFabEventTest.cpp
+++ b/source/test/TestApp/PlayFabEventTest.cpp
@@ -110,14 +110,15 @@ namespace PlayFabUnit
         testContext->Fail(error.GenerateErrorReport());
     }
 
-    void PlayFabEventTest::NonStaticEmitEventCallback(std::shared_ptr<const PlayFab::IPlayFabEvent> /*event*/, std::shared_ptr<const PlayFab::IPlayFabEmitEventResponse> /*response*/)
+    void PlayFabEventTest::NonStaticEmitEventCallback(std::shared_ptr<const PlayFab::IPlayFabEvent> event, std::shared_ptr<const PlayFab::IPlayFabEmitEventResponse> /*response*/)
     {
-        eventTestContext->Pass("Private member called back!");
+        auto pfEvent = std::dynamic_pointer_cast<const TestEvent>(event);
+        pfEvent->testContext->Pass("Private member called back!");
     }
 
     void PlayFabEventTest::EmitEventCallback(std::shared_ptr<const PlayFab::IPlayFabEvent> event, std::shared_ptr<const PlayFab::IPlayFabEmitEventResponse> response)
     {
-        auto pfEvent = std::dynamic_pointer_cast<const PlayFab::PlayFabEvent>(event);
+        auto pfEvent = std::dynamic_pointer_cast<const TestEvent>(event);
         auto pfResponse = std::dynamic_pointer_cast<const PlayFab::PlayFabEmitEventResponse>(response);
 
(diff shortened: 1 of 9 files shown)


source/test/TestApp/PlayFabEventTest.cpp
```diff
@@ -369,9 +362,11 @@ namespace PlayFabUnit
         return api;
     }
 
-    std::unique_ptr<PlayFab::PlayFabEvent> PlayFabEventTest::MakeEvent(PlayFab::PlayFabEventType eventType)
+    std::unique_ptr<PlayFab::PlayFabEvent> PlayFabEventTest::MakeEvent(TestContext& testContext, PlayFab::PlayFabEventType eventType)
     {
-        auto event = std::unique_ptr<PlayFab::PlayFabEvent>(new PlayFab::PlayFabEvent());
+        auto testEvent = new TestEvent();
+        testEvent->testContext = &testContext;
+        auto event = std::unique_ptr<PlayFab::PlayFabEvent>(testEvent);
```
MichaelMackendMS: Simplify this by allocating TestEvent directly into the unique_ptr and then dereference the unique_ptr to set testContext.

source/test/TestApp/PlayFabEventTest.h
```diff
@@ -24,6 +24,13 @@ namespace PlayFabUnit
 {
     struct TestContext;
 
+    // A wrapper around TestEvents, so we can track it back to the test that launched the event
+    class TestEvent : public PlayFab::PlayFabEvent
+    {
+    public:
+        TestContext* testContext;
```
MichaelMackendMS: Raw pointer bad!  But I can see how it has come to this.  If you could at least initialize it to {nullptr} it will look less terrifying.

source/test/TestApp/TestRunner.cpp
```diff
@@ -85,6 +85,7 @@ namespace PlayFabUnit
             test->endTime = PlayFab::GetMilliTicks();
             test->testCase->TearDown(*test);
             test->activeState = TestActiveState::COMPLETE;
+            // printf("\n%s\n", suiteTestSummary.c_str()); // If we're debugging EventTests, it's nice to see each test as it happens...
```
MichaelMackendMS: might be more useful to set a #define for this

source/test/TestApp/PlayFabEventTest.h
pgilmorepf: A future iteration might be able to use weak_ptr<> or something, but that's too big of a change to make right now.

source/test/TestApp/TestRunner.cpp
pgilmorepf: I don't really want to "engineer a solution" for a temporary situation.

source/test/TestApp/PlayFabEventTest.cpp
pgilmorepf: Your suggestion feels like a style issue, and de-ref-ing and re-casting the unique ptr seems a lot uglier than just keeping the ref to the original pointer.

PR #69, opened 2019-12-21
Adding DISABLE_PLAYFAB_STATIC_API

pgilmorepf:
DISABLE_PLAYFAB_STATIC_API can be used to disable all static API interfaces/files to reduce the size of the ExecuteCloudScript
Updating all the features and tests to use instances (with staticPlayer, which is equivalent) internally, rather than static APIs
A handful of header improvements (not nearly complete, but a few steps in the right direction)
Adding some missing AuthenticationContext signatures (IsClientLoggedIn and IsEntityLoggedIn)
Utilizing the context a little better in some tests, rather than copying to internal variables

View the SDK changes that mirror these template changes here:
https://github.com/PlayFab/XPlatCppSdk/pull/38/files

### DIFF
--- a/make.js
+++ b/make.js
@@ -7,9 +7,10 @@ if (typeof templatizeTree === "undefined") templatizeTree = function () { };
 exports.makeCombinedAPI = function (apis, sourceDir, apiOutputDir) {
     console.log("Generating Combined api from: " + sourceDir + " to: " + apiOutputDir);
 
-    var libDefines = "ENABLE_PLAYFABADMIN_API;ENABLE_PLAYFABSERVER_API;";
-    var clientDefines = "";
-    var serverDefines = "ENABLE_PLAYFABADMIN_API;ENABLE_PLAYFABSERVER_API;DISABLE_PLAYFABCLIENT_API;";
+    var removeStatic = ""; // "DISABLE_PLAYFAB_STATIC_API;";
+    var libDefines = "ENABLE_PLAYFABADMIN_API;ENABLE_PLAYFABSERVER_API;" + removeStatic;
+    var clientDefines = "" + removeStatic;
+    var serverDefines = "ENABLE_PLAYFABADMIN_API;ENABLE_PLAYFABSERVER_API;DISABLE_PLAYFABCLIENT_API;" + removeStatic;
 
     var locals = {
         apis: apis,
@@ -31,6 +32,8 @@ exports.makeCombinedAPI = function (apis, sourceDir, apiOutputDir) {
 };
 
 function makeApiFiles(api, sourceDir, apiOutputDir) {
+    var remStaticDefine = ""; // " && !defined(DISABLE_PLAYFAB_STATIC_API)";
+
     var locals = {
         api: api,
         enumtypes: getEnumTypes(api.datatypes),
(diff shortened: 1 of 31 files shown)


make.js
```diff
@@ -31,6 +32,8 @@ exports.makeCombinedAPI = function (apis, sourceDir, apiOutputDir) {
 };
 
 function makeApiFiles(api, sourceDir, apiOutputDir) {
+    var remStaticDefine = ""; // " && !defined(DISABLE_PLAYFAB_STATIC_API)";
```
toddbellMSFT: nit: this looks a bit out of place

source/code/include/playfab/PlayFabSettings.h
```diff
@@ -1,11 +1,13 @@
 #pragma once
 
 #include <playfab/PlayFabError.h>
-#include <playfab/PlayFabSettings.h>
-#include <playfab/PlayFabCallRequestContainer.h>
+//#include <playfab/PlayFabCallRequestContainer.h>
```
toddbellMSFT: stale commented line?

source/code/source/playfab/PlayFabEventPipeline.cpp
```diff
@@ -216,7 +220,7 @@ namespace PlayFab
                 // give some time back to CPU, don't starve it without a good reason
                 std::this_thread::sleep_for(std::chrono::milliseconds(this->settings->readBufferWaitTime));
             }
-            catch (const std::exception& ex)
+            catch (const std::exception & ex)
```
toddbellMSFT: Is this done for an explicit reason? I usually see no space between a reference and the object it refers to. It doesn't change anything according to the compiler, but it's just visual consistency asthetics.

source/test/TestApp/PlayFabApiTest.h
```diff
@@ -45,16 +50,18 @@ namespace PlayFabUnit
...
+        std::shared_ptr<PlayFab::PlayFabDataInstanceAPI> dataApi;
+
         // Fixed values provided from testInputs
         bool TITLE_INFO_SET;
         std::string USER_EMAIL;
 
         // Information fetched by appropriate API calls
         const static std::string TEST_DATA_KEY;
         const static std::string TEST_STAT_NAME;
-        std::string playFabId;
```
toddbellMSFT: Love the cleanup #Resolved

source/code/source/playfab/PlayFabEventPipeline.cpp
pgilmorepf: VS auto spacing default difference between VS 17 & 19
Will fix

---
In reply to: [360618016](https://github.com/PlayFab/XPlatCoreTemplate/pull/69#discussion_r360618016) [](ancestors = 360618016)

make.js
pgilmorepf: This one is deliberate.
If we set remStaticDefine to the commented value, then the define will be applied to the test projects and libraries, and will immediately take effect across the entire solution.
It's basically a single-toggle that enables or disables this feature.
For customers, it's published "enabled" as is the current default.

---
In reply to: [360617787](https://github.com/PlayFab/XPlatCoreTemplate/pull/69#discussion_r360617787) [](ancestors = 360617787)

source/code/include/playfab/PlayFabAuthenticationContext.h
```diff
@@ -9,7 +9,7 @@ namespace PlayFab
     /// </summary>
     class PlayFabAuthenticationContext
     {
-#ifndef DISABLE_PLAYFABCLIENT_API
+#if !defined(DISABLE_PLAYFABCLIENT_API)
```
ScottMunroMS: >#if !defined( [](start = 0, length = 13)

why the change?

PR #70, opened 2020-01-22
Xbox fixes for pch

pgilmorepf:
Some critical includes were moved to PCH files, and the Xbox project wasn't set up properly to use the PCH.

This fixes the project, and the files that don't import the pch properly.

### DIFF
--- a/source/code/include/playfab/PlayFabIXHR2HttpPlugin.h
+++ b/source/code/include/playfab/PlayFabIXHR2HttpPlugin.h
@@ -1,6 +1,5 @@
 #pragma once
 
-#include <stdafx.h>
 #include <playfab/PlayFabCallRequestContainer.h>
 #include <playfab/PlayFabPluginManager.h>
 #include <playfab/PlayFabError.h>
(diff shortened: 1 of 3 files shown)


source/code/include/playfab/PlayFabIXHR2HttpPlugin.h
```diff
@@ -1,5 +1,6 @@
 #pragma once
 
+#include <stdafx.h>
```
ScottMunroMS: I just realized I made these changes in the header. I don't think your other headers do this. my bad. I think this actually belongs in the associated cpp files

ScottMunroMS: figured out why this was needed. it's because of a problem in the xbox vcxproj file again. I'll highlight the problem there for ease

ScottMunroMS:
>     <ClCompile Include="$(SdkSourceDir)\include\playfab\PlayFabIXHR2HttpPlugin.h" />

this should say ClInclude and not ClCompile
same for the next line.

this is why those stdafx.h headers needed to be added to the ixhr2 header files

if you change these two lines to use ClInclude you can revert the diffs in the ixhr2 headers

---
Refers to: source/build/Xbox/XPlatXbox.vcxproj.ejs:188 in ded0d67. [](commit_id = ded0d6760a916dcab3b014e2ca018e4e66dc71aa, deletion_comment = False)

PR #73, opened 2020-01-23
Add if defs for Xbox explicit code

toddbellMSFT:
Customers reported this code wasn't working on a platform that Xbox code won't be able to run in.

### DIFF
--- a/source/code/include/playfab/PlayFabIXHR2HttpPlugin.h
+++ b/source/code/include/playfab/PlayFabIXHR2HttpPlugin.h
@@ -1,5 +1,5 @@
-#if defined(PLAYFAB_PLATFORM_XBOX)
 #pragma once
+#if defined(PLAYFAB_PLATFORM_XBOX)
 
 #include <playfab/PlayFabCallRequestContainer.h>
 #include <playfab/PlayFabPluginManager.h>
--- a/source/code/include/playfab/PlayFabIXHR2HttpPlugin.h
+++ b/source/code/include/playfab/PlayFabIXHR2HttpPlugin.h
@@ -1,3 +1,4 @@
+#if defined(PLAYFAB_PLATFORM_XBOX)
 #pragma once
 
 #include <playfab/PlayFabCallRequestContainer.h>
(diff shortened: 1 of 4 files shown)


source/code/include/playfab/PlayFabIXHR2HttpPlugin.h
```diff
@@ -1,3 +1,4 @@
+#if defined(PLAYFAB_PLATFORM_XBOX)
 #pragma once
```
Vicodex: You may move this pragma higher, out of the protection block

Vicodex:
> #pragma once

You may move pragma out of the block, for consistency

---
Refers to: source/code/include/playfab/PlayFabIXHR2HttpRequest.h:12 in 91650b9. [](commit_id = 91650b9daeaceeea237426d7e258f0a9202abe62, deletion_comment = False)

source/code/source/playfab/PlayFabIXHR2HttpPlugin.cpp
```diff
@@ -1,5 +1,6 @@
 #include <stdafx.h>
```
Vicodex: you may move stdafx include under the protection block

source/code/source/playfab/PlayFabIXHR2HttpRequest.cpp
```diff
@@ -11,6 +11,8 @@
 
 #include <stdafx.h>
```
Vicodex: Same here, you may move it under

source/code/include/playfab/PlayFabIXHR2HttpPlugin.h
pgilmorepf: 👍

source/code/source/playfab/PlayFabIXHR2HttpPlugin.cpp
pgilmorepf: I disagree Vicodex. In this case, you can't. It'll error if #include <stdafx.h> is not present, and the first code line, even if there's nothing else in the .cpp file.

source/code/source/playfab/PlayFabIXHR2HttpRequest.cpp
pgilmorepf: Repeat, I disagree, and it's likely a compile error.

PR #74, opened 2020-01-28
QoS can only properly be enabled for Windows and Xbox

toddbellMSFT:
Explicilty removing enabling flags for QoS in favor of platform specific changes.

Also removes pragma warning disables and fixes the warning issue directly.

### DIFF
--- a/source/test/TestApp/PlayFabEventTest.cpp
+++ b/source/test/TestApp/PlayFabEventTest.cpp
@@ -11,7 +11,9 @@
 #include <playfab/PlayFabEventsDataModels.h>
 #include <playfab/PlayFabEventsInstanceApi.h>
 #include <playfab/PlayFabSettings.h>
+#if QOS_ENABLED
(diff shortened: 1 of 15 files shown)


source/test/TestApp/PlayFabEventTest.cpp
```diff
@@ -11,7 +11,9 @@
 #include <playfab/PlayFabEventsDataModels.h>
 #include <playfab/PlayFabEventsInstanceApi.h>
 #include <playfab/PlayFabSettings.h>
+#if QOS_ENABLED
```
pgilmorepf: I'm pretty tempted to move this entire test to a new file.
It's a bit out of place in "EventTest".
Then we can disable the whole file temporarily, and put a TODO: task number at the top.

toddbellMSFT: Done.

PR #78, opened 2020-02-05
Create set-gitmodules.sh

toddbellMSFT:
adding a bash file so mac development is a bit easier

### DIFF
--- a/source/set-gitmodules.sh
+++ b/source/set-gitmodules.sh
@@ -1,12 +1,14 @@
 #!/bin/bash
 
 #echo ========= Setting Git Submodules to specific commits (BEGIN) =========
-if [ ".git" ] then
+if [ ".git" ]
+then
 	git submodule update --init --recursive
 else
 	git init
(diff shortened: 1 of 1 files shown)


source/set-gitmodules.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
+
+#echo ========= Setting Git Submodules to specific commits (BEGIN) =========
+if [ ".git" ] then
+	git submodule update --init --recursive
+else
+	git init
+	git submodule add https://github.com/open-source-parsers/jsoncpp.git ./external/jsoncpp
+	#if /I "%ERRORLEVEL%" gtr "1" (
```
MichaelMackendMS: why are there commented out error handlers? (here and at 17)

toddbellMSFT: Was a hacky solution to get me unblocked. I have added equivalent bash checks to the script we will commit.

PR #81, opened 2020-02-18
Add necessary missing Android build files

pgilmorepf:
Some Android build files weren't in the template, but should have been.
Adding them fixes the XPlat Android build.
Adding some fields that attempt to filter out the x86 and x64 builds, but it only seems partially successful. While these changes aren't sufficient, I think they are necessary intermediate steps to eventually not-building those binaries, and improving build times.


pgilmorepf:
The only "change" in these files is the addition of:
abiFilters 'armeabi-v7a', 'arm64-v8a'
Everything else is just adding files to the template, exactly as they exist in the output repo.

PR #82, opened 2020-02-20
Adding debug output

pgilmorepf:
Fix a case-sensitivity problem in one of the code files.
Helps diagnose Linux test issues when present.

### DIFF
--- a/source/build/Linux/TestAppLinux.cpp
+++ b/source/build/Linux/TestAppLinux.cpp
@@ -13,10 +13,13 @@ namespace PlayFabUnit
         // Get the path to the test data file from the current environment vars.
         std::string titleDataPath;
         {
-            char* titleDataPathCStr =  getenv("PF_TEST_TITLE_DATA_JSON");
+            char* titleDataPathCStr = getenv("PF_TEST_TITLE_DATA_JSON");
 
             if (titleDataPathCStr == nullptr)
+            {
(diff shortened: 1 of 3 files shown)


source/build/Linux/TestAppLinux.cpp
```diff
@@ -48,6 +52,6 @@ namespace PlayFabUnit
     void TestApp::LogPut(const char* message)
     {
         // Write to STDOUT.
-        puts(message);
```
toddbellMSFT: lol, great catch!

source/build/Linux/TestAppLinux.cpp
```diff
@@ -13,10 +13,13 @@ namespace PlayFabUnit
         // Get the path to the test data file from the current environment vars.
         std::string titleDataPath;
         {
-            char* titleDataPathCStr =  getenv("PF_TEST_TITLE_DATA_JSON");
+            char* titleDataPathCStr = getenv("PF_TEST_TITLE_DATA_JSON");
 
             if (titleDataPathCStr == nullptr)
+            {
+                printf("Could not load PF_TEST_TITLE_DATA_JSON var: %s\n", titleDataPathCStr);
```
toddbellMSFT: Should we consider the secure version of printf? Or if this is a shipped test it doesn't really matter, this will handle any linux terminal printing.

pgilmorepf: TestApp. Not part of the official SDK.
If we can do it in 30 seconds, sure. Otherwise, I don't care enough.

PR #85, opened 2020-03-09
Update XPlatSocket.cpp

toddbellMSFT:
initializing timeOutVal in QoS. As a struct, certain compilers will do fine with this, but GNU has explicit expectations on this time struct being initialized before use.

We supposedly also didn't take into account the micro second nature of _usec according to one customer reported issue (just converting to milliseconds is wrong since usec is alredy noted as micro seconds).

### DIFF
--- a/source/code/source/playfab/QoS/XPlatSocket.cpp
+++ b/source/code/source/playfab/QoS/XPlatSocket.cpp
@@ -20,6 +20,7 @@ namespace PlayFab
         XPlatSocket::XPlatSocket()
         {
             initialized = false;
+            memset(&timeOutVal, 0, sizeof(timeOutVal));
         }
 
         XPlatSocket::~XPlatSocket()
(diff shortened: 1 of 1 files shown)


source/code/source/playfab/QoS/XPlatSocket.cpp
```diff
@@ -20,6 +20,7 @@ namespace PlayFab
         XPlatSocket::XPlatSocket()
         {
             initialized = false;
+            memset(&timeOutVal, 0, sizeof(timeOutVal));
```
pgilmorepf: Please double check the best way to initialize a timreOutVal (whatever it's type is).
It's not necessarily safe to just zero it like this.

pgilmorepf: It could also fail on some platforms and not others.

toddbellMSFT: C-Style casting is the recommendation: https://stackoverflow.com/questions/6462093/reinitialize-timeval-struct

PR #86, opened 2020-03-09
Update PlayFabQoSApi.cpp

toddbellMSFT:
Current implementation can potentially fail in the event of a time out.

A user reported:

"If the future_status status = asyncPingResults[i].wait_for(pingWaitTime); just timeout, the result is never pulled... when the PlayFabQoSApi::GetResult method returns, the vector<future<PingResult>> asyncPingResults variable destruction cause the code to wait for all future tasks not completed yet."

So replacing the initial if with a while in this particular case should be fine.  There is a question though as to whether or not this starts blocking a thread if we are looping in this manner.

### DIFF
--- a/source/code/source/playfab/QoS/PlayFabQoSApi.cpp
+++ b/source/code/source/playfab/QoS/PlayFabQoSApi.cpp
@@ -22,7 +22,7 @@ namespace PlayFab
 {
     namespace QoS
     {
-        static const size_t MaxWaitForFuturesLoopCounts = 1000;
+        static const size_t MaxWaitForFuturesLoopCounts = 300;
         PlayFabQoSApi::PlayFabQoSApi()
         {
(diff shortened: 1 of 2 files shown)


source/code/source/playfab/QoS/PlayFabQoSApi.cpp
```diff
@@ -287,7 +287,7 @@ namespace PlayFab
                 // Iterate over all the threads and servers that need to be pinged
                 for (size_t i = 0; i < numThreads && pingItr < numPings; ++i)
                 {
-                    if (asyncPingResults[i].valid()) // the very first ping result might be a fake future
+                    while (asyncPingResults[i].valid()) // the very first ping result might be a fake future
```
pgilmorepf: I'm not super comfortable with an endless while.
I think it should be a for loop with a fixed-max number of loops so that we don't hang the thread (or process) in this loop.
(const value at the top of the file defining how max-loops used here)

PR #89, opened 2020-03-26
adding a lock to pf event pipeline

toddbellMSFT:
We may read/write to batchesInFlight on multiple threads (either worker thread, callback, or other possible thread if you inherit from our pipeline). Therefore, we should add some locks to this particular structure. 

This bug was created earlier this week: https://dev.azure.com/playfab/PlayFab/_workitems/edit/42545

### DIFF
--- a/source/code/include/playfab/PlayFabEventPipeline.h
+++ b/source/code/include/playfab/PlayFabEventPipeline.h
@@ -73,7 +73,7 @@ namespace PlayFab
         void SetExceptionCallback(ExceptionCallback callback);
 
     protected:
-        virtual void SendBatch(std::vector<std::shared_ptr<const IPlayFabEmitEventRequest>>& batch, size_t& batchCounter);
+        virtual void SendBatch(std::vector<std::shared_ptr<const IPlayFabEmitEventRequest>>& batch, uintptr_t& batchCounter);
 
     private:
         void WorkerThread();
--- a/source/code/include/playfab/PlayFabEventPipeline.h
+++ b/source/code/include/playfab/PlayFabEventPipeline.h
@@ -87,6 +87,7 @@ namespace PlayFab
         // We are using that feature (custom pointer relay) because we need to know which batch it was when we receive a callback from the Events API.
         // To keep track of all batches currently in flight (i.e. those for which we called Events API) we need to have a container with controllable size
         // that would allow to quickly map a pointer (like void* customData) to a batch (like a std::vector<std::shared_ptr<const IPlayFabEmitEventRequest>>).
+        std::mutex inFlightMutex;
         std::unordered_map<void*, std::vector<std::shared_ptr<const IPlayFabEmitEventRequest>>> batchesInFlight;
         std::vector<std::shared_ptr<const IPlayFabEmitEventRequest>> batch;
 
--- a/source/code/include/playfab/PlayFabEventPipeline.h
+++ b/source/code/include/playfab/PlayFabEventPipeline.h
@@ -89,7 +89,6 @@ namespace PlayFab
         // that would allow to quickly map a pointer (like void* customData) to a batch (like a std::vector<std::shared_ptr<const IPlayFabEmitEventRequest>>).
         std::mutex inFlightMutex;
         std::unordered_map<void*, std::vector<std::shared_ptr<const IPlayFabEmitEventRequest>>> batchesInFlight;
-        //std::vector<std::shared_ptr<const IPlayFabEmitEventRequest>> batch;
 
     private:
         std::shared_ptr<PlayFabEventsInstanceAPI> eventsApi;
--- a/source/code/include/playfab/PlayFabEventPipeline.h
+++ b/source/code/include/playfab/PlayFabEventPipeline.h
@@ -73,7 +73,7 @@ namespace PlayFab
         void SetExceptionCallback(ExceptionCallback callback);
 
     protected:
-        virtual void SendBatch(size_t& batchCounter);
+        virtual void SendBatch(std::vector<std::shared_ptr<const IPlayFabEmitEventRequest>>& batch, size_t& batchCounter);
 
     private:
         void WorkerThread();
@@ -89,7 +89,7 @@ namespace PlayFab
         // that would allow to quickly map a pointer (like void* customData) to a batch (like a std::vector<std::shared_ptr<const IPlayFabEmitEventRequest>>).
         std::mutex inFlightMutex;
         std::unordered_map<void*, std::vector<std::shared_ptr<const IPlayFabEmitEventRequest>>> batchesInFlight;
-        std::vector<std::shared_ptr<const IPlayFabEmitEventRequest>> batch;
+        //std::vector<std::shared_ptr<const IPlayFabEmitEventRequest>> batch;
 
     private:
         std::shared_ptr<PlayFabEventsInstanceAPI> eventsApi;
--- a/source/code/include/playfab/PlayFabEventPipeline.h
+++ b/source/code/include/playfab/PlayFabEventPipeline.h
@@ -100,6 +100,8 @@ namespace PlayFab
         std::atomic<bool> isWorkerThreadRunning;
         std::mutex userExceptionCallbackMutex;
         ExceptionCallback userExceptionCallback;
+
+        bool TryGetBatchOutOfFlight(void* customData, std::vector<std::shared_ptr<const IPlayFabEmitEventRequest>>* batchReturn);
     };
 }
 
--- a/source/code/source/playfab/PlayFabEventPipeline.cpp
+++ b/source/code/source/playfab/PlayFabEventPipeline.cpp
@@ -318,7 +318,7 @@ namespace PlayFab
                     playFabEmitEventResponse->playFabError = playFabError;
                     playFabEmitEventResponse->writeEventsResponse = std::shared_ptr<EventsModels::WriteEventsResponse>(new EventsModels::WriteEventsResponse(result));
                     playFabEmitEventResponse->batch = requestBatchPtr;
-                    playFabEmitEventResponse->batchNumber = reinterpret_cast<size_t>(customData);
+                    playFabEmitEventResponse->batchNumber = static_cast<size_t>(reinterpret_cast<uintptr_t>(customData));
 
                     // call an emit event callback
                     CallbackRequest(playFabEmitRequest, std::move(playFabEmitEventResponse));
--- a/source/code/source/playfab/PlayFabEventPipeline.cpp
+++ b/source/code/source/playfab/PlayFabEventPipeline.cpp
@@ -163,11 +163,11 @@ namespace PlayFab
         
         std::vector<std::shared_ptr<const IPlayFabEmitEventRequest>> batch;
 
-        batch.reserve(this->settings->maximalNumberOfItemsInBatch);
-
-        while (this->isWorkerThreadRunning)
+        try
         {
-            try
+            batch.reserve(this->settings->maximalNumberOfItemsInBatch);
+        
+            while (this->isWorkerThreadRunning)
             {
                 size_t sizeOfBatchesInFlight = 0;
 
@@ -230,23 +230,23 @@ namespace PlayFab
                 // give some time back to CPU, don't starve it without a good reason
                 std::this_thread::sleep_for(std::chrono::milliseconds(this->settings->readBufferWaitTime));
             }
-            catch (const std::exception& ex)
-            {
-                LOG_PIPELINE("An exception was caught in PlayFabEventPipeline::WorkerThread method");
-                this->isWorkerThreadRunning = false;
+        }
+        catch (const std::exception& ex)
+        {
+            LOG_PIPELINE("An exception was caught in PlayFabEventPipeline::WorkerThread method");
+            this->isWorkerThreadRunning = false;
 
-                { // LOCK userCallbackMutex
-                    std::unique_lock<std::mutex> lock(userExceptionCallbackMutex);
-                    if (userExceptionCallback)
-                    {
-                        userExceptionCallback(ex);
-                    }
-                } // UNLOCK userCallbackMutex
-            }
-            catch (...)
-            {
-                LOG_PIPELINE("A non std::exception was caught in PlayFabEventPipeline::WorkerThread method");
-            }
+            { // LOCK userCallbackMutex
+                std::unique_lock<std::mutex> lock(userExceptionCallbackMutex);
+                if (userExceptionCallback)
+                {
+                    userExceptionCallback(ex);
+                }
+            } // UNLOCK userCallbackMutex
+        }
+        catch (...)
+        {
+            LOG_PIPELINE("A non std::exception was caught in PlayFabEventPipeline::WorkerThread method");
         }
     }
 
@@ -393,6 +393,7 @@ namespace PlayFab
             // not finding the batch in the queue is a bug
             LOG_PIPELINE("Untracked batch was returned to EventsAPI.WriteEvents callback");
             return false;
+            // UNLOCK batchesInFlight
         }
 
         *batchReturn = std::move(iter->second);
--- a/source/code/source/playfab/PlayFabEventPipeline.cpp
+++ b/source/code/source/playfab/PlayFabEventPipeline.cpp
@@ -45,7 +45,6 @@ namespace PlayFab
         eventsApi = std::make_shared<PlayFabEventsInstanceAPI>(PlayFabSettings::staticPlayer);
 
         this->settings = settings;
-        //this->batch.reserve(this->settings->maximalNumberOfItemsInBatch);
         this->batchesInFlight.reserve(this->settings->maximalNumberOfBatchesInFlight);
         this->Start();
     }
@@ -174,7 +173,7 @@ namespace PlayFab
 
                 { // LOCK batchesInFlight mutex
                     std::unique_lock<std::mutex> lock(inFlightMutex);
-                    sizeOfBatchesInFlight = batchesInFlight.size();
+                    sizeOfBatchesInFlight = this->batchesInFlight.size();
                 } // UNLOCK batchesInFlight
 
                 // Process events in the loop
--- a/source/code/source/playfab/PlayFabEventPipeline.cpp
+++ b/source/code/source/playfab/PlayFabEventPipeline.cpp
@@ -294,27 +294,16 @@ namespace PlayFab
         }
     }
 
+
+
     void PlayFabEventPipeline::WriteEventsApiCallback(const EventsModels::WriteEventsResponse& result, void* customData)
     {
         try
         {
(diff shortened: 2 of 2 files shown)


source/code/source/playfab/PlayFabEventPipeline.cpp
```diff
@@ -286,9 +298,17 @@ namespace PlayFab
...
+            std::unordered_map<void*, std::vector<std::shared_ptr<const IPlayFabEmitEventRequest>>>::iterator foundBatchIterator;
+
+            { // LOCK batchesInFlight mutex
+                std::unique_lock<std::mutex> lock(inFlightMutex);
+                // batch was successfully sent out, find it in the batch tracking map using customData pointer as a key
+                foundBatchIterator = this->batchesInFlight.find(customData);
+                batchMissingFromFlight = foundBatchIterator == this->batchesInFlight.end();
+            } // UNLOCK batchesInFlight
+            
+            if (batchMissingFromFlight)
```
scottnm: instead of checking for the batch under one lock and then removing that batch under another, why not just pop the batch out in one mutually exclusive operation?

```
std::vector<std::shared_ptr<const IPlayFabEmitEventRequest>> batchWritten;
{ // LOCK
    std::unique_lock lock(..);
    auto iter = this->batchesInFlight.find(customData);
    if (iter == this->batchesInFlight.end())
    {
        // this is a bug.
        LOG_PIPELINE(...);
        return;
    }

    batchWritten = std::move(iter->second);
} // UNLOCK

auto requestBatchPtr = ...;
```

this has the benefit of taking one lock and reducing some if-else nesting (a safety AND readability win)

it's usually pretty unsafe (or at least an anti pattern) in multithreaded code to access some shared stated in a mutually exclusive context, keep a record of it, and then use that record later to modify that shared state.

consider the following contrived example + pseudocode
```
int intRecord;
// take mutual exclusive access to the shared integer state
{
    intMutex.lock();
    intRecord = sharedInt.load();
}

// do something outside of the lock

// re-take mutual exclusive access to the shared integer state
if (intRecord < INT32_MAX)
{
    intMutex.lock();
    sharedInt.store(intRecord + 1); // unsafe. another thread could have called into this same function, performed this increment, and now we are overwriting their value.
    // another unsafe idea, sharedInt = sharedInt + 1; 
}
```

source/code/source/playfab/PlayFabEventPipeline.cpp
```diff
@@ -314,8 +334,11 @@ namespace PlayFab
                     CallbackRequest(playFabEmitRequest, std::move(playFabEmitEventResponse));
                 }
 
-                // remove the batch from tracking map
-                this->batchesInFlight.erase(foundBatchIterator->first);
+                { // LOCK batchesInFlight mutex
+                    std::unique_lock<std::mutex> lock2(inFlightMutex);
+                    // remove the batch from tracking map
+                    this->batchesInFlight.erase(foundBatchIterator->first);
```
scottnm: `this->batchesInFlight.erase(iter->first)` is unnecessary.
just use `this->batchesInFlight.erase(iter)` (you avoid having to rehash the key and find the iterator because you already have the iterator).

source/code/source/playfab/PlayFabEventPipeline.cpp
```diff
@@ -328,9 +351,17 @@ namespace PlayFab
...
+            std::unordered_map<void*, std::vector<std::shared_ptr<const IPlayFabEmitEventRequest>>>::iterator foundBatchIterator;
+
+            { // LOCK batchesInFlight mutex
+                std::unique_lock<std::mutex> lock(inFlightMutex);
+                // batch wasn't sent out due to an error, find it in the batch tracking map using customData pointer as a key
+                foundBatchIterator = this->batchesInFlight.find(customData);
+                batchMissingFromFlight = foundBatchIterator == this->batchesInFlight.end();
+            }
+            
+            if (batchMissingFromFlight)
```
scottnm: same comment about not access/writes to this->batchesInFlight over two locked blocks. Collapse into one.

source/code/source/playfab/PlayFabEventPipeline.cpp
ScottMunroMS: instead of checking for the batch under one lock and then removing that batch under another, why not just pop the batch out in one mutually exclusive operation?

```
std::vector<std::shared_ptr<const IPlayFabEmitEventRequest>> batchWritten;
{ // LOCK
    std::unique_lock lock(..);
    auto iter = this->batchesInFlight.find(customData);
    if (iter == this->batchesInFlight.end())
    {
        // this is a bug.
        LOG_PIPELINE(...);
        return;
    }

    batchWritten = std::move(iter->second);
    this->batchesInFlight.erase(iter);
} // UNLOCK

auto requestBatchPtr = ...;
```

this has the benefit of taking one lock instead of two and reducing some if-else nesting (a safety AND readability win)

it's usually pretty unsafe (or at least an anti pattern) in multithreaded code to access some shared stated in a mutually exclusive context, keep a record of it, and then use that record later to modify that shared state.

consider the following contrived example + pseudocode
```
int intRecord;
// take mutual exclusive access to the shared integer state
{
    intMutex.lock();
    intRecord = sharedInt.load();
}

// do something outside of the lock

// re-take mutual exclusive access to the shared integer state
if (intRecord < INT32_MAX)
{
    intMutex.lock();
    sharedInt.store(intRecord + 1); // unsafe. another thread could have called into this same function, performed this increment, and now we are overwriting their value.
    // another unsafe idea, sharedInt = sharedInt + 1; 
}
```

source/code/source/playfab/PlayFabEventPipeline.cpp
ScottMunroMS: `this->batchesInFlight.erase(iter->first)` is unnecessary.
just use `this->batchesInFlight.erase(iter)` (you avoid having to rehash the key and find the iterator because you already have the iterator).

source/code/source/playfab/PlayFabEventPipeline.cpp
ScottMunroMS: same comment about not access/writes to this->batchesInFlight over two locked blocks. Collapse into one.

source/code/include/playfab/PlayFabEventPipeline.h
```diff
@@ -87,6 +87,7 @@ namespace PlayFab
         // We are using that feature (custom pointer relay) because we need to know which batch it was when we receive a callback from the Events API.
         // To keep track of all batches currently in flight (i.e. those for which we called Events API) we need to have a container with controllable size
         // that would allow to quickly map a pointer (like void* customData) to a batch (like a std::vector<std::shared_ptr<const IPlayFabEmitEventRequest>>).
+        std::mutex inFlightMutex;
         std::unordered_map<void*, std::vector<std::shared_ptr<const IPlayFabEmitEventRequest>>> batchesInFlight;
         std::vector<std::shared_ptr<const IPlayFabEmitEventRequest>> batch;
```
ScottMunroMS: I recommend removing `batch` as a member variable and just turning it into a stack variable local to the worker thread.
The easiest way to reduce multi-threading bugs around shared state is to eliminate the possibility of shared state :)

by making this variable local to the worker thread and just passing it into SendBatch (that's the only function that I see using it), you can guarantee that access to it is threadsafe.

```
    void PlayFabEventPipeline::WorkerThread()
    {
        std::vector<std::shared_ptr<const IPlayFabEmitEventRequest>> batch;
        batch.reserve(...);
        // rest of worker thread
```

```
   // have SendBatch accept the worker thread stack batch variable directly instead of referencing it via "this" pointer
    void PlayFabEventPipeline::SendBatch(std::vector<std::shared_ptr<const IPlayFabEmitEventRequest>>& batchsize_t& batchCounter) { ... }
```
```

File: source/code/source/playfab/PlayFabEventPipeline.cpp [hunk 5]
```diff
@@ -257,7 +264,12 @@ namespace PlayFab
 
         // add batch to flight tracking map
         void* customData = reinterpret_cast<void*>(batchCounter); // used to track batches across asynchronous Events API
```
ScottMunroMS: I just noticed this and it's fine to punt it from this review, but I thought I'd bring it up so you could consider addressing it in this change or filing a separate bug...

you reinterpret_cast from `size_t` to `void*` and back.
it's a bit unsafe since `size_t` and `void*` aren't guaranteed to be the same size (which is required for the transform back and forth to be absolutely safe). As a matter of fact this is exactly what the `uinptr_t` type is for! recommend replacing use of `size_t` for the batchCounter with `uintptr_t` one of these days

source/code/source/playfab/PlayFabEventPipeline.cpp
```diff
@@ -166,8 +166,15 @@ namespace PlayFab
...
+                size_t sizeOfBatchesInFlight = 0;
+
+                { // LOCK batchesInFlight mutex
+                    std::unique_lock<std::mutex> lock(inFlightMutex);
+                    sizeOfBatchesInFlight = batchesInFlight.size();
+                } // UNLOCK batchesInFlight
+
                 // Process events in the loop
-                if (this->batchesInFlight.size() >= this->settings->maximalNumberOfBatchesInFlight)
+                if (sizeOfBatchesInFlight >= this->settings->maximalNumberOfBatchesInFlight)
```
ScottMunroMS: do all of the references to `this->settings` need to be reevaluated for thread safety in this file?
It looks like the event pipeline exposes a public `GetSettings` method which returns a mutable shared_ptr to the settings field. That enables external callers to modify the settings used by this event pipeline across any number of threads while the worker thread is running.

This can cause tearing of the state in the `settings` member which could lead to some unexpected behavior if anyone tries to mess with it while the worker thread is running.

unfortunately the `this->settings` value is used to keep track of the auth context needed in the event pipeline so I imagine making it thread safe could prove difficult... (but not impossible!)

source/code/source/playfab/PlayFabEventPipeline.cpp
```diff
@@ -351,9 +382,11 @@ namespace PlayFab
                     // call an emit event callback
                     CallbackRequest(playFabEmitRequest, std::move(playFabEmitEventResponse));
                 }
-
-                // remove the batch from tracking map
-                this->batchesInFlight.erase(foundBatchIterator->first);
+                { // LOCK batchesInFlight mutex
+                    std::unique_lock<std::mutex> lock2(inFlightMutex);
```
jdeweyMSFT: >lock2 [](start = 49, length = 5)

The locks exist in different scopes, so this can be named `lock`

source/code/source/playfab/PlayFabEventPipeline.cpp
```diff
@@ -166,8 +166,15 @@ namespace PlayFab
         {
             try
             {
+                size_t sizeOfBatchesInFlight = 0;
+
+                { // LOCK batchesInFlight mutex
+                    std::unique_lock<std::mutex> lock(inFlightMutex);
+                    sizeOfBatchesInFlight = batchesInFlight.size();
```
jdeweyMSFT: >batchesInFlight [](start = 44, length = 15)

probably needs decorated with the unnecessary `this->` to be consistent with the rest of the code.

source/code/source/playfab/PlayFabEventPipeline.cpp
toddbellMSFT: This change is starting to get big. I have created a different bug to track this particular issue: https://dev.azure.com/playfab/PlayFab/_workitems/edit/42685

ScottMunroMS: that seems super reasonable to me! modifying the settings in the middle of the event api running is probably not an intended calling pattern and it's not a regression so I wouldn't prioritize it for this change.

but I think it's worth tracking as a potential place customers could misuse the api. thanks for filing the bug. I'll resolve this thread.

source/code/include/playfab/PlayFabEventPipeline.h
ScottMunroMS: resolved!

source/code/source/playfab/PlayFabEventPipeline.cpp
ScottMunroMS: there's a `reinterpret_cast<size_t>(customData)` that's still lingering around which probably needs to be changed to `static_cast<size_t>(reinterpret_cast<uintptr_t>(customData))` (the extra static cast is there because the EmitEventResponse still has it's `batch` member stored as a size_t value

source/code/source/playfab/PlayFabEventPipeline.cpp
```diff
@@ -313,10 +323,12 @@ namespace PlayFab
                     // call an emit event callback
                     CallbackRequest(playFabEmitRequest, std::move(playFabEmitEventResponse));
                 }
-
-                // remove the batch from tracking map
-                this->batchesInFlight.erase(foundBatchIterator->first);
             }
+            else
+            {
+                LOG_PIPELINE("After a balid PlayFabEventPipeline write call, the requeted return Batch did not appear in our known flighted batches, there is no trustworthy data we can return to the user");
```
ScottMunroMS: you've got some typos in this LOG_PIPELINE statement. "balid", "requeted"

source/code/source/playfab/PlayFabEventPipeline.cpp
```diff
@@ -376,6 +384,24 @@ namespace PlayFab
...
+        {
+            // not finding the batch in the queue is a bug
+            LOG_PIPELINE("Untracked batch was returned to EventsAPI.WriteEvents callback"); // normally this never happens
+            return false;
+        }
+
+        *batchReturn = std::move(iter->second);
+        this->batchesInFlight.erase(iter);
+        return true;
+        // UNLOCK batchesInFlight
```
ScottMunroMS: I'm not super familiar with how your team likes these UNLOCK comments, but it's noting that you also unlock at the return false line on 397 (which is good I'm just not sure if you were missing a comment)

source/code/source/playfab/PlayFabEventPipeline.cpp
```diff
@@ -159,15 +158,26 @@ namespace PlayFab
         using clock = std::chrono::steady_clock;
         using Result = PlayFabEventBuffer::EventConsumingResult;
         std::shared_ptr<const IPlayFabEmitEventRequest> request;
-        size_t batchCounter = 1; // used to track uniqueness of batches in the map
+        uintptr_t batchCounter = 1; // used to track uniqueness of batches in the map
         std::chrono::steady_clock::time_point momentBatchStarted; // used to track when a currently assembled batch got its first event
+        
+        std::vector<std::shared_ptr<const IPlayFabEmitEventRequest>> batch;
+
+        batch.reserve(this->settings->maximalNumberOfItemsInBatch);
```
ScottMunroMS: this reserve statement can throw. I think that means you want to hoist the `try {}` block so it wraps the entire while loop. You've got an exception block which seems to allow continuing the while loop even if an exception fires (see line 246) but that doesn't seem intentional so just hoisting the try block seems like the right fix

source/code/source/playfab/PlayFabEventPipeline.cpp
```diff
@@ -376,6 +384,24 @@ namespace PlayFab
...
     }
+
+    bool PlayFabEventPipeline::TryGetBatchOutOfFlight(void* customData, std::vector<std::shared_ptr<const IPlayFabEmitEventRequest>>* batchReturn)
+    {
+        // LOCK batchesInFlight mutex
+        std::unique_lock<std::mutex> lock(inFlightMutex);
+        auto iter = this->batchesInFlight.find(customData);
+        if (iter == this->batchesInFlight.end())
+        {
+            // not finding the batch in the queue is a bug
```
ScottMunroMS: I think having the `// not find the batch in the queue is a bug` and `// normal this never happens` comments below is redundant. I would opt for keeping the former comment and removing the latter comment since the first comment is clear that something is a bug rather than uncommon but expected behavior.

source/code/source/playfab/PlayFabEventPipeline.cpp
```diff
@@ -328,15 +340,10 @@ namespace PlayFab
...
-            if (foundBatchIterator == this->batchesInFlight.end())
-            {
-                LOG_PIPELINE("Untracked batch was returned to EventsAPI.WriteEvents callback"); // normally this never happens
-            }
-            else
+            std::vector<std::shared_ptr<const IPlayFabEmitEventRequest>> batchWritten;
+            if(TryGetBatchOutOfFlight(customData, &batchWritten))
             {
-                auto requestBatchPtr = std::shared_ptr<const std::vector<std::shared_ptr<const IPlayFabEmitEventRequest>>>(new std::vector<std::shared_ptr<const IPlayFabEmitEventRequest>>(std::move(foundBatchIterator->second)));
+                auto requestBatchPtr = std::shared_ptr<const std::vector<std::shared_ptr<const IPlayFabEmitEventRequest>>>(new std::vector<std::shared_ptr<const IPlayFabEmitEventRequest>>(batchWritten));
```
ScottMunroMS: you lost std::move when declaring these requestBatchPtr values. I think you still want to keep those or else you will copy the entire batchWritten vector unnecessarily

source/code/source/playfab/PlayFabEventPipeline.cpp
```diff
@@ -286,15 +301,10 @@ namespace PlayFab
...
```
ScottMunroMS: these requestBatchPtr lines are a doozy to read.
you can make them shorter (by like 60 chars) if you use std::make_shared

```
auto requestBatchPtr = std::make_shared<std::vector<std::shared_ptr<const IPlayFabEmitEventRequest>>>(std::move(batchWritten));
```

source/code/source/playfab/PlayFabEventPipeline.cpp
ScottMunroMS: sorry looks like there is actually a second one to fix in the error callback as well

source/code/source/playfab/PlayFabEventPipeline.cpp
toddbellMSFT: I thought I was moving the std::move to the batchReturn. But it still makes sense to be moving the object if it is movable.

ScottMunroMS: you are correct, you are std::move-ing into batchReturn
but batchWritten still needs to be std::moved if you want to avoid the copy when you construct the new shared_ptr variable. Shared_ptr is an owning pointer which means it won't take a weak reference to batchReturn. It will copy unless you use std::move

there shouldn't be a need to clear out `batchWritten` after you've std::moved it though

source/code/source/playfab/PlayFabEventPipeline.cpp
toddbellMSFT: since PlayFabEmitEventReponse is already shipped, it may break existing customers if we change the signature to be size_t, so I think we have to keep the static_casts you provided.

Issue #79, opened 2020-02-06
Intermittent crash in ExecuteRequest during shutdown


MSFT-Chris-Barrett:
One of our top iOS crashes looks to be in block execution within ExecuteRequest().  So far I've managed to repro it only once in a debugger, and it was this line due to `body` being nil: https://github.com/PlayFab/XPlatCoreTemplate/blob/ded594258c367d3e92909f951d826854f3a4dfab/source/code/source/playfab/PlayFabIOSHttpPlugin.mm#L271

All parameters given when calling the block were nil: `data`, `response`, and `error`.  This crash occurs _sometimes_ when a user is closing our app or it's otherwise exiting.  In the one case I repro'd, it was from dismissing/swiping-up the app from the overview of running apps.

Current guess is toward the app exiting while a request is completing.  So _something somewhere_ is destroyed, then this completionHandler block is run.  I'm still learning Objective-C for the first time to look into this, so I'm not yet familiar with what-all cases to watch out for in how relevant memory-is/objects-are being handled.

pgilmorepf:
Thanks for reporting this, I'll file the issue with the team.

toddbellMSFT:
This issue has been addressed with the latest commit: 

https://github.com/PlayFab/XPlatCoreTemplate/commit/93e00bec4ec9cbf75ff0f35e630b4d09a3dade02 

Please let us know if you continue to see this issue after the next release.

PR #91, opened 2020-03-30
Update PlayFabIOSHttpPlugin.mm

toddbellMSFT:
Quick fix to see if the object named body is nil before we attempt to reference it in a response back to a user to fix this customer reported issue

https://github.com/PlayFab/XPlatCoreTemplate/issues/79

If PlayFab fail's here, we will be sending back an error to the user. So the responseString should reflect the error to be more easily debugged by the customer (if the body is nil, there would have been no information to give back to the customer anyway, so adding in error info to this response would be more helpful than no info.

### DIFF
--- a/source/code/source/playfab/PlayFabIOSHttpPlugin.mm
+++ b/source/code/source/playfab/PlayFabIOSHttpPlugin.mm
@@ -272,10 +272,6 @@ namespace PlayFab
                                                                        {
                                                                            requestContainer.responseString = [body UTF8String];
                                                                        }
-                                                                       else
-                                                                       {
-                                                                           requestContainer.responseString = "PlayFabIOSHttpPlugin failed to parse body as it is nil";
-                                                                       }
                                                                        ProcessResponse(*request, static_cast<const int>(httpResponse.statusCode));
                                                                        { // LOCK httpRequestMutex
                                                                            lock->lock();
--- a/source/code/source/playfab/PlayFabIOSHttpPlugin.mm
+++ b/source/code/source/playfab/PlayFabIOSHttpPlugin.mm
@@ -272,6 +272,10 @@ namespace PlayFab
                                                                        {
                                                                         requestContainer.responseString = [body UTF8String];
                                                                        }
+                                                                       else
+                                                                       {
+                                                                           requestContainer.responseString = "PlayFabIOSHttpPlugin Failed to parse body";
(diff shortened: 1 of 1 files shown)


source/code/source/playfab/PlayFabIOSHttpPlugin.mm
```diff
@@ -268,7 +268,14 @@ void Cancel()
                                                                        NSString* body = [[NSString alloc] initWithData:data encoding:NSUTF8StringEncoding];
 
                                                                        CallRequestContainer& requestContainer = request->RequestContainer();
-                                                                       requestContainer.responseString = [body UTF8String];
+                                                                       if(body != nil)
+                                                                       {
+                                                                        requestContainer.responseString = [body UTF8String];
```
Vicodex: Please keep indentation consistent

source/code/source/playfab/PlayFabIOSHttpPlugin.mm
```diff
@@ -268,7 +268,14 @@ void Cancel()
...
+                                                                       }
+                                                                       else
+                                                                       {
+                                                                           requestContainer.responseString = "PlayFabIOSHttpPlugin Failed to parse body";
```
Vicodex: I think something like "PlayFabIOSHttpPlugin Failed to parse body as it is nil" would be even more informative

source/code/source/playfab/PlayFabIOSHttpPlugin.mm
```diff
@@ -268,7 +268,14 @@ void Cancel()
...
 
                                                                        CallRequestContainer& requestContainer = request->RequestContainer();
-                                                                       requestContainer.responseString = [body UTF8String];
+                                                                       if(body != nil)
+                                                                       {
+                                                                           requestContainer.responseString = [body UTF8String];
+                                                                       }
+                                                                       else
+                                                                       {
+                                                                           requestContainer.responseString = "PlayFabIOSHttpPlugin failed to parse body as it is nil";
```
Toddsa: We should not be setting a response string to a non empty or nil value. Do we know why it is crashing, is the length in addition to it being nil?

Toddsa: To be more specific, adding in our hint to the response is not generally acceptable.

PR #92, opened 2020-04-07
Toddbell/adding settings comment

toddbellMSFT:
There are some known threading issues with some variables in our EventPipeline. We don't have the resources to dedicate refactoring this codebase to make it more threadsafe, so we are making a quick note about how the settings should be expected to work.

### DIFF
--- a/source/code/include/playfab/PlayFabEventPipeline.h
+++ b/source/code/include/playfab/PlayFabEventPipeline.h
@@ -66,9 +66,9 @@ namespace PlayFab
         PlayFabEventPipeline& operator=(const PlayFabEventPipeline& source) = delete; // disable assignment
         PlayFabEventPipeline& operator=(PlayFabEventPipeline&& other) = delete; // disable move assignment
 
-        // NOTE: settings are expected to be set before calling ::Start()
-        // after PlayFabEventPipeline::Start() has been called, 
-        // it is recommended you no longer change the settings.
+        // NOTE: settings are expected to be set prior to calling PlayFabEventPipeline::Start()
+        // changing them after PlayFabEventPipeline::Start() may cause threading issues
+        // users should not expect changes made to settings to take effect after ::Start is called unless the pipeline is destroyed and re-created
         std::shared_ptr<PlayFabEventPipelineSettings> GetSettings() const;
         virtual void Start() override;
         virtual void IntakeEvent(std::shared_ptr<const IPlayFabEmitEventRequest> request) override;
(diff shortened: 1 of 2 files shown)


source/code/include/playfab/PlayFabEventPipeline.h
```diff
@@ -66,6 +66,9 @@ namespace PlayFab
         PlayFabEventPipeline& operator=(const PlayFabEventPipeline& source) = delete; // disable assignment
         PlayFabEventPipeline& operator=(PlayFabEventPipeline&& other) = delete; // disable move assignment
 
+        // NOTE: settings are expected to be set prior to calling PlayFabEventPipeline::Start()
+        // changing them after PlayFabEventPipeline::Start() may cause threading issues
+        // users should not expect changes made to settings to take effect after ::Start is called unless the pipeline is destroyed and re-created
```
ScottMunroMS, line 71: users can't stop the pipeline so after it starts they basically HAVE to destroy and recreate if they want to change anything

what about introducing the minor breaking change of just having GetSettings return a std::shared_ptr<const PlayFabEventPipelineSettings>`. It's more correct and better implies what you are trying to communicate with a comment. I also think it's fine because it's a breaking change but only in a codepath/feature which was already broken/not supported

ScottMunroMS: edit: whoops my last comment had some malformed markdown in it so it wasn't readable :( apologies and fixed :)

toddbellMSFT: We are going to avoid any breaking changes to the public facing APIs at this time, no matter how minor. With the current situation in mind, we explicitly shouldn't be adding any more work for our customers, especially if it can function as originally intended. 

We could possibly add in a ::Stop mechanism to help circumvent the complete destroy/recreate issue, but that feature request may take some time before we have the bandwidth to get to it.

ScottMunroMS: I'd like to reiterate that this will only cause "more work" for customers that are using the API in a way that isn't safe and that is likely the tiniest subset of customers. For the small set of customers impacted, the breaks would be trivial to resolve unless they were doing something unsafe in which case it would be good to break them so they can fix their unsafe usage.

ScottMunroMS: and then customers don't have to go around looking for comments among a huge list of changes in the next release to see if they need to update anything. I think the emphasis on not breaking customers is actually hurting customers in this case.

I'm going to remove my PR block but I think currently this PR is addressing the situation in a suboptimal way

PR #93, opened 2020-04-21

toddbellMSFT:
check for HRESULT when adding headers. WinHttpAddRequestHeaders can potentially fail in a few ways (out of memory for example). We don't handle this kind of issue at the moment.

### DIFF
--- a/source/code/include/playfab/PlayFabWinHttpPlugin.h
+++ b/source/code/include/playfab/PlayFabWinHttpPlugin.h
@@ -36,8 +36,8 @@ namespace PlayFab
         void HandleResults(std::unique_ptr<CallRequestContainer> requestContainer);
         void SetErrorInfo(CallRequestContainer& requestContainer, const std::string& errorMessage, const int httpCode = 0) const;
 
-        bool TryAddHeader(HINTERNET hRequest, LPCWSTR lpszHeaders, DWORD dwHeadersLength, DWORD dwModifiers);
-        void CloseOutConnections(std::unique_ptr<CallRequestContainer> requestContainer, HINTERNET hRequest, HINTERNET hConnect, HINTERNET hSession);
+        bool TryAddHeader(HINTERNET hRequest, LPCWSTR lpszHeaders);
+        void CompleteRequest(std::unique_ptr<CallRequestContainer> requestContainer, HINTERNET hRequest, HINTERNET hConnect, HINTERNET hSession);
 
         std::thread workerThread;
         std::mutex httpRequestMutex;
--- a/source/code/include/playfab/PlayFabWinHttpPlugin.h
+++ b/source/code/include/playfab/PlayFabWinHttpPlugin.h
@@ -42,7 +42,7 @@ namespace PlayFab
         std::thread workerThread;
         std::mutex httpRequestMutex;
         std::atomic<bool> threadRunning;
-        std::atomic<bool> headerNoErrors;
+        std::atomic<bool> setPredefinedHeadersFailed;
         int activeRequestCount;
         std::deque<std::unique_ptr<CallRequestContainerBase>> pendingRequests;
         std::deque<std::unique_ptr<CallRequestContainerBase>> pendingResults;
--- a/source/code/include/playfab/PlayFabWinHttpPlugin.h
+++ b/source/code/include/playfab/PlayFabWinHttpPlugin.h
@@ -42,7 +42,7 @@ namespace PlayFab
         std::thread workerThread;
         std::mutex httpRequestMutex;
         std::atomic<bool> threadRunning;
-        std::atomic<BOOL> setPredefinedHeadersFailed;
+        std::atomic<HRESULT> setPredefinedHeadersResult;
         int activeRequestCount;
         std::deque<std::unique_ptr<CallRequestContainerBase>> pendingRequests;
         std::deque<std::unique_ptr<CallRequestContainerBase>> pendingResults;
--- a/source/code/include/playfab/PlayFabWinHttpPlugin.h
+++ b/source/code/include/playfab/PlayFabWinHttpPlugin.h
@@ -36,13 +36,13 @@ namespace PlayFab
         void HandleResults(std::unique_ptr<CallRequestContainer> requestContainer);
         void SetErrorInfo(CallRequestContainer& requestContainer, const std::string& errorMessage, const int httpCode = 0) const;
 
-        bool TryAddHeader(HINTERNET hRequest, LPCWSTR lpszHeaders);
+        BOOL TryAddHeader(HINTERNET hRequest, LPCWSTR lpszHeaders);
         void CompleteRequest(std::unique_ptr<CallRequestContainer> requestContainer, HINTERNET hRequest, HINTERNET hConnect, HINTERNET hSession);
 
         std::thread workerThread;
         std::mutex httpRequestMutex;
         std::atomic<bool> threadRunning;
-        std::atomic<bool> setPredefinedHeadersFailed;
+        std::atomic<BOOL> setPredefinedHeadersFailed;
         int activeRequestCount;
         std::deque<std::unique_ptr<CallRequestContainerBase>> pendingRequests;
         std::deque<std::unique_ptr<CallRequestContainerBase>> pendingResults;
--- a/source/code/include/playfab/PlayFabWinHttpPlugin.h
+++ b/source/code/include/playfab/PlayFabWinHttpPlugin.h
@@ -36,9 +36,13 @@ namespace PlayFab
         void HandleResults(std::unique_ptr<CallRequestContainer> requestContainer);
         void SetErrorInfo(CallRequestContainer& requestContainer, const std::string& errorMessage, const int httpCode = 0) const;
 
+        bool TryAddHeader(HINTERNET hRequest, LPCWSTR lpszHeaders, DWORD dwHeadersLength, DWORD dwModifiers);
+        void CloseOutConnections(std::unique_ptr<CallRequestContainer> requestContainer, HINTERNET hRequest, HINTERNET hConnect, HINTERNET hSession);
+
         std::thread workerThread;
         std::mutex httpRequestMutex;
         std::atomic<bool> threadRunning;
+        std::atomic<bool> headerNoErrors;
         int activeRequestCount;
         std::deque<std::unique_ptr<CallRequestContainerBase>> pendingRequests;
         std::deque<std::unique_ptr<CallRequestContainerBase>> pendingResults;
--- a/source/code/include/playfab/PlayFabWinHttpPlugin.h
+++ b/source/code/include/playfab/PlayFabWinHttpPlugin.h
@@ -36,7 +36,7 @@ namespace PlayFab
         void HandleResults(std::unique_ptr<CallRequestContainer> requestContainer);
         void SetErrorInfo(CallRequestContainer& requestContainer, const std::string& errorMessage, const int httpCode = 0) const;
 
-        BOOL TryAddHeader(HINTERNET hRequest, LPCWSTR lpszHeaders);
+        HRESULT TryAddHeader(HINTERNET hRequest, LPCWSTR lpszHeaders);
         void CompleteRequest(std::unique_ptr<CallRequestContainer> requestContainer, HINTERNET hRequest, HINTERNET hConnect, HINTERNET hSession);
 
         std::thread workerThread;
--- a/source/code/source/playfab/PlayFabWinHttpPlugin.cpp
+++ b/source/code/source/playfab/PlayFabWinHttpPlugin.cpp
@@ -16,7 +16,7 @@ namespace PlayFab
     {
         activeRequestCount = 0;
         threadRunning = true;
-        setPredefinedHeadersFailed = false;
+        setPredefinedHeadersFailed = FALSE;
         workerThread = std::thread(&PlayFabWinHttpPlugin::WorkerThread, this);
     };
 
@@ -376,9 +376,9 @@ namespace PlayFab
             TryAddHeader(hRequest, L"X-ReportErrorAsSuccess: true");
     }
 
-    bool PlayFabWinHttpPlugin::TryAddHeader(HINTERNET hRequest, LPCWSTR lpszHeaders)
+    BOOL PlayFabWinHttpPlugin::TryAddHeader(HINTERNET hRequest, LPCWSTR lpszHeaders)
     {
-        return static_cast<bool>(WinHttpAddRequestHeaders(hRequest, lpszHeaders, -1, 0));
+        return WinHttpAddRequestHeaders(hRequest, lpszHeaders, -1, 0);
     }
 
     bool PlayFabWinHttpPlugin::GetBinaryPayload(CallRequestContainer& reqContainer, LPVOID& payload, DWORD& payloadSize) const
--- a/source/code/source/playfab/PlayFabWinHttpPlugin.cpp
+++ b/source/code/source/playfab/PlayFabWinHttpPlugin.cpp
@@ -193,9 +193,10 @@ namespace PlayFab
                     {
                         // Add HTTP headers
                         SetPredefinedHeaders(reqContainer, hRequest);
-                        if(setPredefinedHeadersResult != S_OK)
+                        HRESULT hr = setPredefinedHeadersResult.exchange(S_OK);
+                        if(FAILED(hr))
                         {
-                            SetErrorInfo(reqContainer, "Error in attempting to add Default Headers with HRESULT: " + std::to_string(setPredefinedHeadersResult));
+                            SetErrorInfo(reqContainer, "Error in attempting to add Default Headers with HRESULT: " + std::to_string(hr));
                             CompleteRequest(std::move(requestContainer), hRequest, hConnect, hSession);
                             setPredefinedHeadersResult = S_OK;
                             return;
@@ -209,9 +210,10 @@ namespace PlayFab
                                 if (obj.first.length() != 0 && obj.second.length() != 0) // no empty keys or values in headers
                                 {
                                     std::string header = obj.first + ": " + obj.second;
-                                    if(!TryAddHeader(hRequest, std::wstring(header.begin(), header.end()).c_str()))
+                                    hr = TryAddHeader(hRequest, std::wstring(header.begin(), header.end()).c_str());
+                                    if(FAILED(hr))
                                     {
-                                        SetErrorInfo(reqContainer, "Error in WinHttpAddRequestHeaders attempting to add parameters with HRESULT: " + std::to_string(GetLastError()));
+                                        SetErrorInfo(reqContainer, "Error in WinHttpAddRequestHeaders attempting to add parameters with HRESULT: " + std::to_string(hr));
                                         CompleteRequest(std::move(requestContainer), hRequest, hConnect, hSession);
                                         return;
                                     }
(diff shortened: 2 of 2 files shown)


source/code/source/playfab/PlayFabWinHttpPlugin.cpp
```diff
@@ -200,7 +200,11 @@ namespace PlayFab
                                 if (obj.first.length() != 0 && obj.second.length() != 0) // no empty keys or values in headers
                                 {
                                     std::string header = obj.first + ": " + obj.second;
-                                    WinHttpAddRequestHeaders(hRequest, std::wstring(header.begin(), header.end()).c_str(), -1, 0);
+                                    HRESULT hHeaderAdd = WinHttpAddRequestHeaders(hRequest, std::wstring(header.begin(), header.end()).c_str(), -1, 0);
+                                    if(!hHeaderAdd)
+                                    {
+                                        SetErrorInfo(reqContainer, "Error in WinHttpAddRequestHeaders");
```
ScottMunroMS, line 206: this will set the error info, but won't it still fire off the request? won't the response from that request overwrite the error info? it doesn't look like anything here is actually failing this loop early

source/code/source/playfab/PlayFabWinHttpPlugin.cpp
```diff
@@ -200,7 +200,11 @@ namespace PlayFab
                                 if (obj.first.length() != 0 && obj.second.length() != 0) // no empty keys or values in headers
                                 {
                                     std::string header = obj.first + ": " + obj.second;
-                                    WinHttpAddRequestHeaders(hRequest, std::wstring(header.begin(), header.end()).c_str(), -1, 0);
+                                    HRESULT hHeaderAdd = WinHttpAddRequestHeaders(hRequest, std::wstring(header.begin(), header.end()).c_str(), -1, 0);
```
ScottMunroMS, line 203: I did a quick scan of this file and saw other instances of WinHttpAddRequestHeaders which were not being checked for failures (SetPredefinedHeaders or something)

source/code/source/playfab/PlayFabWinHttpPlugin.cpp
toddbellMSFT: I can add a break after the SetErrorInfo to cut out early

ScottMunroMS: I'll leave it to you to fix it, but I think just adding a break after seterrorinfo will only break out of this "add headers" loop which still won't stop the http request from being sent off

ScottMunroMS:
how was this change tested? I'm not sure if it warrants a new test (though I certainly wouldn't fight that ;) ), but knowing what measure of success was used to verify the fix would give me confidence that this change hasn't introduced some odd side effect during this new failure path

pgilmorepf:
retest

pgilmorepf:
retest

pgilmorepf:
retest

source/code/source/playfab/PlayFabWinHttpPlugin.cpp
toddbellMSFT: Yeah, I noticed the break would only break out one loop. I opted to instead do a return from the function, and created a helper that would close any open connections we may have made (moved the end part of the ExecuteRequest function to this new helper function and call it in 3 places, first 2 will set an error before hand and then return right after, and the final one is the correct happy path to complete the ExecuteRequest.

source/code/source/playfab/PlayFabWinHttpPlugin.cpp
```diff
@@ -192,6 +193,14 @@ namespace PlayFab
                     {
                         // Add HTTP headers
                         SetPredefinedHeaders(reqContainer, hRequest);
+                        if(!headerNoErrors)
```
ScottMunroMS, line 196: why is SetPredefinedHeaders returning void and relying on coupling with a member variable to report errors? this would be much more straight forward if SetPredefinedHeaders just returned the HRESULT

```
HRESULT hr = SetPredefinedHeaders(...);
if (FAILED(hr))
{
    ....
}
```

source/code/source/playfab/PlayFabWinHttpPlugin.cpp
```diff
@@ -372,10 +370,15 @@ namespace PlayFab
...
-        WinHttpAddRequestHeaders(hRequest, L"X-ReportErrorAsSuccess: true", -1, 0);
+        headerNoErrors = !TryAddHeader(hRequest, L"Accept: application/json", -1, 0) &&
+            !TryAddHeader(hRequest, L"Content-Type: application/json; charset=utf-8", -1, 0) &&
+            !TryAddHeader(hRequest, (L"X-PlayFabSDK: " + std::wstring(PlayFabSettings::versionString.begin(), PlayFabSettings::versionString.end())).c_str(), -1, 0) &&
+            !TryAddHeader(hRequest, L"X-ReportErrorAsSuccess: true", -1, 0);
+    }
+
+    bool PlayFabWinHttpPlugin::TryAddHeader(HINTERNET hRequest, LPCWSTR lpszHeaders, DWORD dwHeadersLength, DWORD dwModifiers)
+    {
+        return WinHttpAddRequestHeaders(hRequest, lpszHeaders, dwHeadersLength, dwModifiers) == S_OK;
```
ScottMunroMS, line 381: WinHttpAddRequestHeaders does not return an HRESULT. [It returns a bool](https://docs.microsoft.com/en-us/windows/win32/api/winhttp/nf-winhttp-winhttpaddrequestheaders).

ScottMunroMS: which means any time we successfully add headers this will return false (since S_OK is 0). which means I think this breaks even the golden path success case

source/code/source/playfab/PlayFabWinHttpPlugin.cpp
ScottMunroMS: Additionally, if you return an hresult from SetPredefinedHeaders you can just use the return value on the SetErrorInfo line below instead of calling GetLastError which isn't as clear IMO

source/code/source/playfab/PlayFabWinHttpPlugin.cpp
```diff
@@ -200,7 +209,12 @@ namespace PlayFab
                                 if (obj.first.length() != 0 && obj.second.length() != 0) // no empty keys or values in headers
                                 {
                                     std::string header = obj.first + ": " + obj.second;
-                                    WinHttpAddRequestHeaders(hRequest, std::wstring(header.begin(), header.end()).c_str(), -1, 0);
+                                    if(TryAddHeader(hRequest, std::wstring(header.begin(), header.end()).c_str(), -1, 0))
+                                    {
+                                        SetErrorInfo(reqContainer, "Error in WinHttpAddRequestHeaders attempting to add parameters" + std::to_string(GetLastError()));
```
ScottMunroMS, line 214: nit: this error message won't have any space between it and the hresult value

it also should probably mention that the value being reported is an HRSEULT

source/code/source/playfab/PlayFabWinHttpPlugin.cpp
```diff
@@ -466,6 +469,27 @@ namespace PlayFab
             return activeRequestCount;
         }
     }
+
+    void PlayFabWinHttpPlugin::CloseOutConnections(std::unique_ptr<CallRequestContainer> requestContainer, HINTERNET hRequest, HINTERNET hConnect, HINTERNET hSession)
```
ScottMunroMS, line 473: nit: the name `CloseOutConnections` feels like it's missing the really important part of what this function does. Handling the callback is what's most important for logically following this code. what about calling this `CompleteRequest` to mirror the `ExecuteRequest where it's called.

source/code/source/playfab/PlayFabWinHttpPlugin.cpp
```diff
@@ -372,10 +370,15 @@ namespace PlayFab
...
-        WinHttpAddRequestHeaders(hRequest, L"Content-Type: application/json; charset=utf-8", -1, 0);
-        WinHttpAddRequestHeaders(hRequest, (L"X-PlayFabSDK: " + std::wstring(PlayFabSettings::versionString.begin(), PlayFabSettings::versionString.end())).c_str(), -1, 0);
-        WinHttpAddRequestHeaders(hRequest, L"X-ReportErrorAsSuccess: true", -1, 0);
+        headerNoErrors = !TryAddHeader(hRequest, L"Accept: application/json", -1, 0) &&
+            !TryAddHeader(hRequest, L"Content-Type: application/json; charset=utf-8", -1, 0) &&
+            !TryAddHeader(hRequest, (L"X-PlayFabSDK: " + std::wstring(PlayFabSettings::versionString.begin(), PlayFabSettings::versionString.end())).c_str(), -1, 0) &&
+            !TryAddHeader(hRequest, L"X-ReportErrorAsSuccess: true", -1, 0);
+    }
+
+    bool PlayFabWinHttpPlugin::TryAddHeader(HINTERNET hRequest, LPCWSTR lpszHeaders, DWORD dwHeadersLength, DWORD dwModifiers)
```
ScottMunroMS, line 379: it looks like all callers of TryAddHeader pass -1 for `dwHeadersLength` and 0 for `dwModifiers`

you could just hardcode those values inside of the TryAddHeader helper until someone actually needs to pass something different. One of the advantages of wrapping external APIs with helper methods is being able to simplify the parameter lists they take (which makes the code calling that wrapper easier to read)

source/code/source/playfab/PlayFabWinHttpPlugin.cpp
```diff
@@ -192,6 +193,14 @@ namespace PlayFab
                     {
                         // Add HTTP headers
                         SetPredefinedHeaders(reqContainer, hRequest);
+                        if(!headerNoErrors)
+                        {
+                            SetErrorInfo(reqContainer, "Error in attempting to add Default Headers with HRESULT: " + std::to_string(GetLastError()));
+                            CloseOutConnections(std::move(requestContainer), hRequest, hConnect, hSession);
```
ScottMunroMS, line 199: nit: it makes it difficult to read that some of the codepaths for `ExecuteRequest` return early on failure whereas other failure codepaths just bubble down to the final CloseOutConnections call by virtue of nested if-else blocks.

it doesn't help that ExecuteRequest is a large function with lots of nesting.

recommend restructuring ExecuteRequest to either use early returns for all failure paths or to not use them at all.

Frankly, I think it would be a general readability improvement to use early returns for all failure cases, but I'll leave that to you and the SDK teams preference (maybe y'all like the nesting. some members of my team sure do). There are a number of ways to enable early returns for all codepaths. You could just explicitly call CloseOutConnections at each failure point or you could turn ExecuteRequest into a wrapper that calls two helper functions, SendRequest and CompleteRequest

```
void ExecuteRequest(...)
{
     HINTERNET hSession = NULL, hConnect = NULL, hRequest = NULL;
     TrySendRequest(*callRequestContainer, &hSession, &hConnect, &hRequest);
     CompleteRequest(std::move(callRequestContainer), hSession, hConnect, hRequest));
}

void
TrySendRequest(CallRequestContainer& reqContainer, HINTERNET* phSession, HINTERNET* phConnect, HINTERNET* phRequest)
{
    bool res = SomeOperation(..);
    if (!res)
    {
        SetErrorInfo(blah blah balh);
        return;
    }
    /// ....
}

void
CompleteRequest(std::unique_ptr<CallRequestContainer> reqContainer, HINTERNET hSession, HINTERNET hConnect, HINTERNET hRequest)
{
   // .. this is just your CloseOutConnection helpers which I've renamed for this context
}
```

toddbellMSFT:
Moving the changes between XPlat back to the SDKGenerator seems to have this issue. There was significant testing involved but notice no Continuous Integration still on this repo?

---
In reply to: [397847910](https://github.com/PlayFab/XPlatCoreTemplate/pull/93#issuecomment-397847910) [](ancestors = 397847910)

source/code/source/playfab/PlayFabWinHttpPlugin.cpp
toddbellMSFT: We are not changing any current method signatures. 

I stated in the updated commit comment that this function call is inherited by all other HTTPPlugins. This particular task only covers WinHttp issues. 

Yes I agree it's a better design to update the function signature to return the underlying HRESULT (trust me it looked WAY better to handle errors inside that function rather than relying on this flag), but since HttpPlugin is public and this function is virtual, this would be a breaking change to users.

---
In reply to: [412662871](https://github.com/PlayFab/XPlatCoreTemplate/pull/93#discussion_r412662871) [](ancestors = 412662871)

ScottMunroMS: oh I think I see. you've designed the API such that interested parties can inherit from the winhttpplugin. oof that's unfortunate.

Can you at least update the variable name to something easier to read than `headerNoErrors` that more clearly identifies the state as being coupled to SetPredefinedHeaders? e.g. `setPreDefinedHeadersFailure` or `setPredefinedHeadersResult` or something like that.

additionally, I think having SetPredefinedHeaders use a std::atomic<HRESULT> instead of std::atomic<bool> which requires calling GetLastError is still goodness

toddbellMSFT: This is not correct. I saw the function in the Windows plugin was set to virtual, but it really has no need of being that way, 

---
In reply to: [413170386](https://github.com/PlayFab/XPlatCoreTemplate/pull/93#discussion_r413170386) [](ancestors = 413170386,412662871)

ScottMunroMS: aren't all of the helper methods (and internal state) of the winhttp plugin marked protected? I think in theory that means a consumer could inherit from the winhttp plugin and take a dependency on the SetPredefinedHeaders method, no? I don't think that's a good idea and if it wasn't an intentionally designed use case maybe it's ok to break, but as is I think it's still possible for consumers to take weird dependencies on the winhttpplugin.

toddbellMSFT: I'm setting to won't fix this particular design decision to preserve whatever original state the code is currently in (focusing on a "keeping the lights on" approach here, rather than re-wiring)

---
In reply to: [413186169](https://github.com/PlayFab/XPlatCoreTemplate/pull/93#discussion_r413186169) [](ancestors = 413186169)

source/code/source/playfab/PlayFabWinHttpPlugin.cpp
```diff
@@ -372,10 +370,15 @@ namespace PlayFab
     void PlayFabWinHttpPlugin::SetPredefinedHeaders(const CallRequestContainer& requestContainer, HINTERNET hRequest)
     {
         UNREFERENCED_PARAMETER(requestContainer);
-        WinHttpAddRequestHeaders(hRequest, L"Accept: application/json", -1, 0);
-        WinHttpAddRequestHeaders(hRequest, L"Content-Type: application/json; charset=utf-8", -1, 0);
-        WinHttpAddRequestHeaders(hRequest, (L"X-PlayFabSDK: " + std::wstring(PlayFabSettings::versionString.begin(), PlayFabSettings::versionString.end())).c_str(), -1, 0);
-        WinHttpAddRequestHeaders(hRequest, L"X-ReportErrorAsSuccess: true", -1, 0);
+        setPredefinedHeadersFailed = TryAddHeader(hRequest, L"Accept: application/json") &&
```
ScottMunroMS, line 373: >setPredefinedHeadersFailed [](start = 8, length = 26)

isn't this backwards? TryAddHeader returns TRUE on success.

when all of the TryAddHeader calls succeed you are setting `setPredefinedHeadersFailed` to TRUE which would indicate a failure, not success

source/code/source/playfab/PlayFabWinHttpPlugin.cpp
```diff
@@ -372,10 +370,15 @@ namespace PlayFab
...
+            TryAddHeader(hRequest, L"Content-Type: application/json; charset=utf-8") &&
+            TryAddHeader(hRequest, (L"X-PlayFabSDK: " + std::wstring(PlayFabSettings::versionString.begin(), PlayFabSettings::versionString.end())).c_str()) &&
+            TryAddHeader(hRequest, L"X-ReportErrorAsSuccess: true");
+    }
+
+    BOOL PlayFabWinHttpPlugin::TryAddHeader(HINTERNET hRequest, LPCWSTR lpszHeaders)
```
ScottMunroMS, line 379: >BOOL [](start = 4, length = 4)

nit: I prefered this as a normal `bool`. putting a dependency on win data types in your public headers (even for a windows specific plugin) is unnecessary

source/code/source/playfab/PlayFabWinHttpPlugin.cpp
```diff
@@ -192,6 +193,14 @@ namespace PlayFab
                     {
                         // Add HTTP headers
                         SetPredefinedHeaders(reqContainer, hRequest);
+                        if(!setPredefinedHeadersFailed)
+                        {
```
ScottMunroMS, line 197: note this if-expression also reads backwards. `if (!failed) { /* failure case */ }`

you should be checking `if(setPredefinedHeadersFailed) { /* failure case */ }` but SetPredefinedHeaders is returning a backwards result

source/code/source/playfab/PlayFabWinHttpPlugin.cpp
```diff
@@ -192,6 +193,14 @@ namespace PlayFab
...
+                            SetErrorInfo(reqContainer, "Error in attempting to add Default Headers with HRESULT: " + std::to_string(GetLastError()));
```
ScottMunroMS, line 198: >HRESULT: " + std::to_string(GetLastError() [](start = 104, length = 42)

the return of GetLastError isn't an HRESULT. It's a win32 system error code. You need to use HRESULT_FROM_WIN32 to convert to an HR

there's another line where you have this same issue below.

source/code/include/playfab/PlayFabWinHttpPlugin.h
```diff
@@ -36,9 +36,13 @@ namespace PlayFab
         void HandleResults(std::unique_ptr<CallRequestContainer> requestContainer);
         void SetErrorInfo(CallRequestContainer& requestContainer, const std::string& errorMessage, const int httpCode = 0) const;
 
+        BOOL TryAddHeader(HINTERNET hRequest, LPCWSTR lpszHeaders);
+        void CompleteRequest(std::unique_ptr<CallRequestContainer> requestContainer, HINTERNET hRequest, HINTERNET hConnect, HINTERNET hSession);
+
         std::thread workerThread;
         std::mutex httpRequestMutex;
         std::atomic<bool> threadRunning;
+        std::atomic<BOOL> setPredefinedHeadersFailed;
```
ScottMunroMS, line 45: >BOOL [](start = 20, length = 4)

I still think a better choice for this value is std::atomic<HRESULT>

if you do that then TryAddHeader can return HRESULTs and can locally do the conversion from WinHttp's bool return to a more synchronous HRESULT and I think that will noticeably improve readability

Here are some examples of where the code changes and improves IMO
```
// ...
// Add HTTP headers
SetPredefinedHeaders(reqContainer, hRequest);
HRESULT hr = setPredefinedHeadersResult.exchange(S_OK);
if(FAILED(hr))
{
    SetErrorInfo(reqContainer, "Error in attempting to add Default Headers with HRESULT: " + std::to_string(hr));
    CompleteRequest(std::move(requestContainer), hRequest, hConnect, hSession);
    return;
}
// ..

// ..
std::string header = obj.first + ": " + obj.second;
HRESULT hr = TryAddHeader(hRequest, std::wstring(header.begin(), header.end()).c_str());
if(FAILED(hr))
{
    SetErrorInfo(reqContainer, "Error in WinHttpAddRequestHeaders attempting to add parameters with HRESULT: " + std::to_string(hr));
    CompleteRequest(std::move(requestContainer), hRequest, hConnect, hSession);
    return;
}
// ...

void PlayFabWinHttpPlugin::SetPredefinedHeaders(const CallRequestContainer& requestContainer, HINTERNET hRequest)
{
    UNREFERENCED_PARAMETER(requestContainer);
    
    HRESULT hr = S_OK;

    hr = TryAddHeader(hRequest, L"Accept: application/json");
    if (SUCCEEDED(hr))
    {
        hr = TryAddHeader(hRequest, L"Content-Type: application/json; charset=utf-8");
        if (SUCCEEDED(hr))
        {
            std::string versionHeader = "X-PlayFabSDK: " + PlayFabSettings::versionString();
            std::wstring versionWideHeader(versionHeader.begin(), versionHeader.end());
            hr = TryAddHeader(hRequest, versionWideHeader.c_str());
            if (SUCCEEDED(hr))
            {
                hr = TryAddHeader(hRequest, L"X-ReportErrorAsSuccess: true");
            }
        }
    }
    
    setPredefinedHeadersResult = hr;
}

HRESULT PlayFabWinHttpPlugin::TryAddHeader(HINTERNET hRequest, LPCWSTR lpszHeaders)
{
    bool succeeded = WinHttpAddRequestHeaders(hRequest, lpszHeaders, -1, 0);
    if (!succeeded)
    {
        return HRESULT_FROM_WIN32(GetLastError());
    }
    return S_OK;
}
```

source/code/source/playfab/PlayFabWinHttpPlugin.cpp
```diff
@@ -192,6 +193,14 @@ namespace PlayFab
...
+                            CompleteRequest(std::move(requestContainer), hRequest, hConnect, hSession);
+                            setPredefinedHeadersFailed = true;
```
ScottMunroMS, line 200: >setPredefinedHeadersFailed [](start = 28, length = 26)

since you are using std::atomic, I recommend taking advantage of that and using a single `exchange` operation instead of splitting the load and store across two lines of code (which defeats part of the point of using an atomic)

```
SetPredefinedHeaders(..);
bool failed = setPredefinedHeadersFailed.exchange(true); // load the value that SetPredefinedHeaders set and atomically clear it back to true
if (!failed)
{
   SetErrorInfo(...)
   CompleteRequest(...);
}
```

 (note I've left comments elsewhere about how the setPredefinedHeadersFailed is backwards and I suspect you actually want to be exchanging to `false` and comparing the if-check for `true` values, but I've left that alone for this comment to demonstrate what would actually change by using `exchange` instead of a separate load and store)

source/code/source/playfab/PlayFabWinHttpPlugin.cpp
```diff
@@ -372,10 +370,18 @@ namespace PlayFab
...
-        WinHttpAddRequestHeaders(hRequest, L"Accept: application/json", -1, 0);
-        WinHttpAddRequestHeaders(hRequest, L"Content-Type: application/json; charset=utf-8", -1, 0);
-        WinHttpAddRequestHeaders(hRequest, (L"X-PlayFabSDK: " + std::wstring(PlayFabSettings::versionString.begin(), PlayFabSettings::versionString.end())).c_str(), -1, 0);
-        WinHttpAddRequestHeaders(hRequest, L"X-ReportErrorAsSuccess: true", -1, 0);
+        if(!TryAddHeader(hRequest, L"Accept: application/json") ||
+            !TryAddHeader(hRequest, L"Content-Type: application/json; charset=utf-8") ||
+            !TryAddHeader(hRequest, (L"X-PlayFabSDK: " + std::wstring(PlayFabSettings::versionString.begin(), PlayFabSettings::versionString.end())).c_str()) ||
+            !TryAddHeader(hRequest, L"X-ReportErrorAsSuccess: true"))
+        {
+            setPredefinedHeadersResult = GetLastError();
```
ScottMunroMS, line 378: you are setting this variable with a win32 system error and not an hresult. You need to use HRESULT_FROM_WIN32 to convert it to an hresult

toddbellMSFT: That came in with the successive change to TryAddHeader (the HRESULT_FROM_WIN32 is called in there)

---
In reply to: [414066496](https://github.com/PlayFab/XPlatCoreTemplate/pull/93#discussion_r414066496) [](ancestors = 414066496)

PR #95, opened 2020-05-06
Toddbell/update declare comments

toddbellMSFT:
Adding a quick comment about general Api Update()

### DIFF
--- a/templates/PlayFab_API.h.ejs
+++ b/templates/PlayFab_API.h.ejs
@@ -18,6 +18,10 @@ namespace PlayFab
     public:
         /// <summary>
         /// Runs the IHttpPlugin's Update function to check for responses to HTTP requests (can be platform specific HttpPlugin impl or user defiend)
+        /// This has global consequences since all api's share the same httpPlugin (technically, call order of Update per Api may matter to developers)
+        /// You should call it in this manner: 
+        ///    <%- api.name %>.Update();
+        /// whenever your own game ticks.
         /// </summary>
         static size_t Update();
         static void ForgetAllCredentials();
--- a/templates/PlayFab_API.h.ejs
+++ b/templates/PlayFab_API.h.ejs
@@ -20,7 +20,7 @@ namespace PlayFab
         /// Calls the Update function on your implementation of the IHttpPlugin to check for responses to HTTP requests.
         /// All api's (Client, Server, Admin etc.) share the same IHttpPlugin. 
         /// This means that you only need to call Update() on one API to retrieve the responses for all APIs.
-        /// Additional calls to Update (on any API) during the same tick is unlikely retrieve addtional responses.
+        /// Additional calls to Update (on any API) during the same tick are unlikely retrieve additional responses.
         /// Call Update when your game ticks as follows:
         ///     <%- api.name %>.Update();
         /// </summary>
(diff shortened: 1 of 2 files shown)


templates/PlayFab_API.h.ejs
```diff
@@ -16,6 +16,16 @@ namespace PlayFab
...
+        /// <summary>
+        /// Runs the IHttpPlugin's Update function to check for responses to HTTP requests (could be a platform specific HttpPluginImpl or user defiend)
+        /// This has global consequences since all api's share the same httpPlugin 
+        /// Therefore, only 1 api needs to call Update() 
+        /// It Doesn't matter which API, all will do the exact same thing.
+        /// Please note any later Update call in the same tick will probably be a wasted call
+        /// You may call it in this manner: 
+        ///    <%- api.name %>.Update();
+        /// whenever your own game ticks.
+        /// </summary>
```
vDonGlover, line 28: You reference IHttpPlugin here, but in subsequent places you reference it in different ways so it is not entirely clear to me what is going on. 
    If I were to guess it might go like this.

    Calls the Update function on your implementation of the IHttpPlugin to check for responses to HTTP requests.
    
    In the asside you reference HttpPluginImpl then later you reference httpPlugin (and IHttpPlugin early) how do they all interact.
    
    This has global consequences since all api's share the same httpPlugin  
    Therefore, only 1 api needs to call Update().
    It Doesn't matter which API, all will do the exact same thing.
    
    Is this actually 
    All api's (maybe call out client, server, admin) share the same httpPlugin. 
    This means that you only need to call Update() on one API to retrieve the responses for all APIs.
        
    Additional cals to Update (on any API) during the same tick is unlikely retrieve addtional responses.
    
    Call Update when your game ticks as follows:
    
        <%- api.name %>.Update();

templates/PlayFab_API.h.ejs
```diff
@@ -17,14 +17,12 @@ namespace PlayFab
...
-        /// Therefore, only 1 api needs to call Update() 
-        /// It Doesn't matter which API, all will do the exact same thing.
-        /// Please note any later Update call in the same tick will probably be a wasted call
-        /// You may call it in this manner: 
-        ///    <%- api.name %>.Update();
-        /// whenever your own game ticks.
+        /// Calls the Update function on your implementation of the IHttpPlugin to check for responses to HTTP requests.
+        /// All api's (Client, Server, Admin etc.) share the same IHttpPlugin. 
+        /// This means that you only need to call Update() on one API to retrieve the responses for all APIs.
+        /// Additional calls to Update (on any API) during the same tick is unlikely retrieve addtional responses.
```
vDonGlover, line 23: I see I made two typos: is should be are and additional was missing an i.
Additional calls to Update (on any API) during the same tick are unlikely retrieve additional responses.

templates/PlayFab_API.h.ejs
```diff
@@ -20,7 +20,7 @@ namespace PlayFab
         /// Calls the Update function on your implementation of the IHttpPlugin to check for responses to HTTP requests.
         /// All api's (Client, Server, Admin etc.) share the same IHttpPlugin. 
         /// This means that you only need to call Update() on one API to retrieve the responses for all APIs.
-        /// Additional calls to Update (on any API) during the same tick is unlikely retrieve addtional responses.
+        /// Additional calls to Update (on any API) during the same tick are unlikely retrieve additional responses.
```
vDonGlover, line 23: unlikely to retrieve additional responses.

PR #96, opened 2020-05-06
Toddbell/event pipeline update

toddbellMSFT:
There was a PR awhile back to a forked version of our project that seemed to be a good fit to port to our public codebase. This PR contains that fix. 

@ScottMunroMS was able to add a threaded callback flag to the event pipeline, enabling manual pumping functionality.

Scott's summary:

"this change updates the playfab event pipeline to allow manually pumping it's work queues instead of relying on over-active background threads.

doing so eliminates extra threads which we couldn't affinitize. Having fewer background threads and ensuring all of them are affinitizable will make AAA devs happy."

### DIFF
--- a/source/code/include/playfab/PlayFabEventApi.h
+++ b/source/code/include/playfab/PlayFabEventApi.h
@@ -29,6 +29,10 @@ namespace PlayFab
 
         void EmitEvent(std::unique_ptr<const IPlayFabEvent> event, std::function<void(std::shared_ptr<const IPlayFabEvent>, std::shared_ptr<const IPlayFabEmitEventResponse>)> callback) const;
 
+        /// <summary>
+        /// Updates the underlying event router which in turn will update the eventpipeline.
+        /// This function must be called every game tick if threadedEventPipeline is set to false
+        /// </summary>
         void Update();
 
     private:
--- a/source/code/include/playfab/PlayFabEventApi.h
+++ b/source/code/include/playfab/PlayFabEventApi.h
@@ -14,9 +14,7 @@ namespace PlayFab
     class PlayFabEventAPI
     {
     public:
-        PlayFabEventAPI(); // Default constructor
-
-        PlayFabEventAPI(bool threadedEventPipeline);
+        PlayFabEventAPI(bool threadedEventPipeline=true);
 
         std::shared_ptr<IPlayFabEventRouter> GetEventRouter() const;
 
--- a/source/code/include/playfab/PlayFabEventApi.h
+++ b/source/code/include/playfab/PlayFabEventApi.h
@@ -15,6 +15,9 @@ namespace PlayFab
     {
     public:
         PlayFabEventAPI(); // Default constructor
+
+        PlayFabEventAPI(bool threadedEventPipeline);
+
         std::shared_ptr<IPlayFabEventRouter> GetEventRouter() const;
 
         /// <summary>
@@ -25,6 +28,8 @@ namespace PlayFab
         void EmitEvent(std::unique_ptr<const IPlayFabEvent> event, const PlayFabEmitEventCallback callback) const;
 
         void EmitEvent(std::unique_ptr<const IPlayFabEvent> event, std::function<void(std::shared_ptr<const IPlayFabEvent>, std::shared_ptr<const IPlayFabEmitEventResponse>)> callback) const;
+
+        void Update();
 
     private:
         std::shared_ptr<IPlayFabEventRouter> eventRouter;
--- a/source/code/include/playfab/PlayFabEventPipeline.h
+++ b/source/code/include/playfab/PlayFabEventPipeline.h
@@ -74,8 +74,9 @@ namespace PlayFab
         PlayFabEventPipeline& operator=(PlayFabEventPipeline&& other) = delete; // disable move assignment
 
         // NOTE: settings are expected to be set prior to calling PlayFabEventPipeline::Start()
-        // changing them after PlayFabEventPipeline::Start() may cause threading issues
-        // users should not expect changes made to settings to take effect after ::Start is called unless the pipeline is destroyed and re-created
+        // changing them after PlayFabEventPipeline::Start() may cause threading issues unless you have set useBackgroundThread flag to true
+        // If this flag is not set, users should not expect changes made to settings to take effect after ::Start is called
+        //   unless the pipeline is A.) destroyed and re-created or B.) restart it by running Stop() and then Start() again 
         std::shared_ptr<PlayFabEventPipelineSettings> GetSettings() const;
         virtual void Start() override;
         virtual void Stop() override;
@@ -102,7 +103,6 @@ namespace PlayFab
         // that would allow to quickly map a pointer (like void* customData) to a batch (like a std::vector<std::shared_ptr<const IPlayFabEmitEventRequest>>).
         std::mutex inFlightMutex;
         std::unordered_map<void*, std::vector<std::shared_ptr<const IPlayFabEmitEventRequest>>> batchesInFlight;
-        //std::vector<std::shared_ptr<const IPlayFabEmitEventRequest>> batch;
 
     private:
         std::shared_ptr<PlayFabEventsInstanceAPI> eventsApi;
--- a/source/code/include/playfab/PlayFabEventPipeline.h
+++ b/source/code/include/playfab/PlayFabEventPipeline.h
@@ -104,10 +104,6 @@ namespace PlayFab
     private:
         std::atomic_uintptr_t batchCounter;
         std::chrono::steady_clock::time_point momentBatchStarted;
-        std::shared_ptr<PlayFabEventsInstanceAPI> eventsApi;
-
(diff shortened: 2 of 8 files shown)


source/code/include/playfab/PlayFabEventApi.h
```diff
@@ -15,6 +15,9 @@ namespace PlayFab
     {
     public:
         PlayFabEventAPI(); // Default constructor
```
MichaelMackendMS, line 17: It is unclear to me what the default behavior is.  The new ctor takes a bool that would seemingly dictate the underlying behavior of the class.  The default ctor appears does what, then?  Should we just have the new ctor with a default parameter?

source/code/include/playfab/PlayFabEventApi.h
```diff
@@ -26,6 +29,8 @@ namespace PlayFab
 
         void EmitEvent(std::unique_ptr<const IPlayFabEvent> event, std::function<void(std::shared_ptr<const IPlayFabEvent>, std::shared_ptr<const IPlayFabEmitEventResponse>)> callback) const;
 
+        void Update();
```
MichaelMackendMS, line 32: Update() is now only functional when the API is not threaded.   This should at least be documented someplace.

source/code/source/playfab/PlayFabEventPipeline.cpp
```diff
@@ -23,33 +23,54 @@ namespace PlayFab
...
+            this->Start();
+        }
     }
 
     PlayFabEventPipeline::~PlayFabEventPipeline()
+    {
+        Stop();
+    }
+
+    void PlayFabEventPipeline::Start()
```
MichaelMackendMS, line 63: This method is public.  It starts the background thread regardless of whether the settings indicate that this is a background threaded event pipeline.  It should check and either except or log error when called incorrectly.

source/code/source/playfab/PlayFabEventPipeline.cpp
```diff
@@ -59,14 +80,18 @@ namespace PlayFab
...
-        // start worker thread
-        this->isWorkerThreadRunning = true;
-        if (!this->workerThread.joinable())
+        if (this->settings->useBackgroundThread)
         {
-            this->workerThread = std::thread(&PlayFabEventPipeline::WorkerThread, this);
+            throw std::runtime_error("You should not call Update() when PlayFabEventPipelineSettings::useBackgroundThread == true");
         }
+
+        bool hasMoreWorkToProcess;
```
MichaelMackendMS, line 90: nit: initialize

source/code/include/playfab/PlayFabEventPipeline.h
```diff
@@ -74,15 +78,18 @@ namespace PlayFab
         // users should not expect changes made to settings to take effect after ::Start is called unless the pipeline is destroyed and re-created
```
MichaelMackendMS, line 78: These comments need to be updated to reflect the correct usage of the class (ie the difference in usage when using threads or not)

source/code/include/playfab/PlayFabEventPipeline.h
```diff
@@ -95,10 +102,13 @@ namespace PlayFab
         // that would allow to quickly map a pointer (like void* customData) to a batch (like a std::vector<std::shared_ptr<const IPlayFabEmitEventRequest>>).
         std::mutex inFlightMutex;
         std::unordered_map<void*, std::vector<std::shared_ptr<const IPlayFabEmitEventRequest>>> batchesInFlight;
+        //std::vector<std::shared_ptr<const IPlayFabEmitEventRequest>> batch;
```
MichaelMackendMS, line 105: we should probably remove this.

source/code/source/playfab/PlayFabEventPipeline.cpp
```diff
@@ -23,33 +23,54 @@ namespace PlayFab
...
 
     PlayFabEventPipeline::~PlayFabEventPipeline()
+    {
+        Stop();
+    }
+
+    void PlayFabEventPipeline::Start()
+    {
+        // start worker thread
+        this->isWorkerThreadRunning = true;
```
MichaelMackendMS, line 66: This function should check if isWorkerThreadRunning is already true and do error handling.

source/code/include/playfab/PlayFabEventRouter.h
```diff
@@ -37,8 +38,10 @@ namespace PlayFab
     class PlayFabEventRouter : public IPlayFabEventRouter
     {
     public:
-        PlayFabEventRouter();
+        PlayFabEventRouter(bool threadedEventPipeline);
         virtual void RouteEvent(std::shared_ptr<const IPlayFabEmitEventRequest> request) const;
+
+        virtual void Update() override;
```
MichaelMackendMS, line 44: Needs comments indicating that this function should only be called when threadedEventPipeline is false.

source/code/include/playfab/PlayFabEventApi.h
toddbellMSFT: I was initially thinking that too, Scott's implementation passes true directly in the cpp constructor implementation, so I opted for 1:1 as much as I could (there were a few places where we made changes that may not have made it back to bumblelion).

source/code/include/playfab/PlayFabEventPipeline.h
toddbellMSFT: We were able to remove batch as a member to help potential threading issues on our codebase by making it more narrowly scoped (as a parameter in SendBatch, which in Scott's review was empty)

source/code/include/playfab/PlayFabEventApi.h
ScottMunroMS: I don't think I'd mind a ctor with a default parameter if that makes this more clear for y'all
`PlayFabEventAPI(bool threadedEventPipeline=true);`

The reason I made the change without one was to try and keep things additive and our team just general tries to avoid default parameters. Also since y'all expose the source it's not difficult to figure out what the default ctor does (though I agree with the sentiment that looking at source files to figure out this behavior is subpar)

source/code/source/playfab/PlayFabEventApi.cpp
```diff
@@ -7,10 +7,16 @@
...
     PlayFabEventAPI::PlayFabEventAPI() :
-        eventRouter(std::make_shared<PlayFabEventRouter>()) // default event router
+        PlayFabEventAPI(true)
     {
     }
 
+    PlayFabEventAPI::PlayFabEventAPI(bool threadedEventPipeline) : 
+        eventRouter(std::make_shared<PlayFabEventRouter>(threadedEventPipeline))
+    {
+
```
ScottMunroMS, line 17: nit: unnecessary whitespace line in ctor body

source/code/include/playfab/PlayFabEventApi.h
ScottMunroMS: though I do appreciate you keeping the change as 1:1 as possible :)

source/code/source/playfab/PlayFabEventPipeline.cpp
```diff
@@ -23,34 +23,73 @@ namespace PlayFab
...
+        if (!this->workerThread.joinable())
+        {
+            this->workerThread = std::thread(&PlayFabEventPipeline::WorkerThread, this);
+        }
+    }
+
+    void PlayFabEventPipeline::Stop()
+    {
+        if(!isWorkerThreadRunning)
```
ScottMunroMS, line 87: should there be a similar if check for if the settings don't specify `useBackgroundThread`?

source/code/source/playfab/PlayFabEventPipeline.cpp
```diff
@@ -184,10 +238,17 @@ namespace PlayFab
...
                     // do not take new events from buffer if batches currently in flight are at the maximum allowed number
                     // and are not sent out (or received an error) yet
-                    std::this_thread::sleep_for(std::chrono::milliseconds(this->settings->readBufferWaitTime)); // give some time for batches in flight to deflate
-                    continue;
+                    return false;
                 }
 
+                // Don't try taking a request until we get an entity token. We'd rather not lose events
+                // that were generated before an entity token was created (i.e. before a user was created).
+                if (PlayFabSettings::staticPlayer->entityToken.empty() &&
```
ScottMunroMS, line 246: I'm torn on this if-block in your change.
in general, I think it's good since..
1. for playfab sdk consumers it enables users of the event pipeline to queue events BEFORE a user logs in
2. it's one less things that's specific to our private fork of the SDK

but I'd be remiss if I didn't speak up and say it's actually not a part of the change to the event pipeline. This is a change that was checked into our internal private fork of the SDK a while ago and was not brought back to the upstream SDK. I think it *should* be brought back in this change (given my reasons above) and I think the only reason it wasn't originally was because a dev likely felt time pressure, but just fyi

source/code/include/playfab/PlayFabEventPipeline.h
MichaelMackendMS: I guess I just meant we should remove this commented out line. :)

PR #97, opened 2020-05-06
Toddbell/header handle xplat

toddbellMSFT:
Curl http plugin is missing header-adding error-handling logic (if we add a header and run out of memory, we need to gracefully handle the issue and stop attempting to send out a broken request).

### DIFF
--- a/source/code/source/playfab/PlayFabCurlHttpPlugin.cpp
+++ b/source/code/source/playfab/PlayFabCurlHttpPlugin.cpp
@@ -333,7 +333,7 @@ namespace PlayFab
         list = curl_slist_append(list, headerToAppend);
         if(list == NULL)
         {
-            CurlHeaderFailed(std::move(requestContainer));
+            CurlHeaderFailed(requestContainer);
         }
         return list;
     }
--- a/source/code/source/playfab/PlayFabCurlHttpPlugin.cpp
+++ b/source/code/source/playfab/PlayFabCurlHttpPlugin.cpp
@@ -298,7 +298,7 @@ namespace PlayFab
         }
     }
 
-    curl_slist* PlayFabCurlHttpPlugin::SetPredefinedHeaders()
+    curl_slist* PlayFabCurlHttpPlugin::SetPredefinedHeaders(CallRequestContainer& reqContainer)
     {
         curl_slist* curlHttpHeaders = nullptr;
 
--- a/source/code/source/playfab/PlayFabCurlHttpPlugin.cpp
+++ b/source/code/source/playfab/PlayFabCurlHttpPlugin.cpp
@@ -325,7 +325,7 @@ namespace PlayFab
         HandleCallback(std::move(requestContainer));
     }
 
-    bool PlayFabCurlHttpPlugin::TryCurlAddHeader(std::unique_ptr<CallRequestContainer> requestContainer, curl_slist* list, char* headerToAppend)
+    bool PlayFabCurlHttpPlugin::TryCurlAddHeader(std::unique_ptr<CallRequestContainer> requestContainer, curl_slist* list, const char* headerToAppend)
     {
         list = curl_slist_append(list, headerToAppend);
         if(list == NULL)
--- a/source/code/source/playfab/PlayFabCurlHttpPlugin.cpp
+++ b/source/code/source/playfab/PlayFabCurlHttpPlugin.cpp
@@ -179,6 +179,15 @@ namespace PlayFab
                 {
                     std::string header = obj.first + ": " + obj.second;
                     curlHttpHeaders = curl_slist_append(curlHttpHeaders, header.c_str());
+                    if(curlHttpHeaders == NULL)
+                    {
+                        reqContainer.errorWrapper.HttpStatus = "Failed to create Headers list";
+                        reqContainer.errorWrapper.ErrorCode = PlayFabErrorCode::UnknownError;
+                        reqContainer.errorWrapper.ErrorName = "Header Creation failed";
+                        reqContainer.errorWrapper.ErrorMessage = "Request failed initializing before sending. Failing out early.";
+                        HandleCallback(std::move(requestContainer));
+                        return;
+                    }
                 }
             }
         }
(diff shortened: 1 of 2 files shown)


source/code/source/playfab/PlayFabCurlHttpPlugin.cpp
```diff
@@ -164,31 +164,31 @@ namespace PlayFab
...
+        }
 
-        curlHttpHeaders = curl_slist_append(curlHttpHeaders, "Accept: application/json");
-        curlHttpHeaders = curl_slist_append(curlHttpHeaders, "Content-Type: application/json; charset=utf-8");
-        curlHttpHeaders = curl_slist_append(curlHttpHeaders, ("X-PlayFabSDK: " + PlayFabSettings::versionString).c_str());
-        curlHttpHeaders = curl_slist_append(curlHttpHeaders, "X-ReportErrorAsSuccess: true");
+        // curlHttpHeaders = curl_slist_append(curlHttpHeaders, "Accept: application/json");
+        // curlHttpHeaders = curl_slist_append(curlHttpHeaders, "Content-Type: application/json; charset=utf-8");
+        // curlHttpHeaders = curl_slist_append(curlHttpHeaders, ("X-PlayFabSDK: " + PlayFabSettings::versionString).c_str());
+        // curlHttpHeaders = curl_slist_append(curlHttpHeaders, "X-ReportErrorAsSuccess: true");
```
pgilmorepf, line 191: Let's do a full test pass across all platforms, and be really certain about this change.
After that, let's remove these code comments, as we've clearly already been down this road.
VERY rigorous testing is needed.

source/code/source/playfab/PlayFabCurlHttpPlugin.cpp
```diff
@@ -291,4 +319,22 @@ namespace PlayFab
...
+    {
+        requestContainer.errorWrapper.HttpStatus = "Failed to create Headers list";
+        requestContainer.errorWrapper.ErrorCode = PlayFabErrorCode::PlayFabErrorUnkownError;
+        requestContainer.errorWrapper.ErrorName = "Header Creation failed";
+        requestContainer.errorWrapper.ErrorMessage = "Request failed initializing before sending. Failing out early.";
+    }
+
+    curl_slist* PlayFabCurlHttpPlugin::TryCurlAddHeader(CallRequestContainer& requestContainer, curl_slist* list, const char* headerToAppend)
+    {
+        list = curl_slist_append(list, headerToAppend);
```
pgilmorepf, line 333:
```suggestion
        if (list != nullptr)
        {
            list = curl_slist_append(list, headerToAppend);
        }
```

source/code/source/playfab/PlayFabCurlHttpPlugin.cpp
```diff
@@ -291,4 +319,22 @@ namespace PlayFab
...
+        if(list == NULL)
+        {
+            CurlHeaderFailed(requestContainer);
```
pgilmorepf, line 336:
```suggestion
```

source/code/source/playfab/PlayFabCurlHttpPlugin.cpp
```diff
@@ -164,10 +163,34 @@ namespace PlayFab
...
+        if (curlHttpHeaders == NULL)
+        {
+            HandleCallback(std::move(requestContainer));
+            return;
+        }
+
+        curlHttpHeaders = TryCurlAddHeader(reqContainer, curlHttpHeaders, "X-ReportErrorAsSuccess: true");
+        if (curlHttpHeaders == NULL)
+        {
+            HandleCallback(std::move(requestContainer));
```
pgilmorepf, line 191:
```suggestion
            CurlHeaderFailed(requestContainer);
            HandleCallback(std::move(requestContainer));
```

source/code/source/playfab/PlayFabCurlHttpPlugin.cpp
```diff
@@ -164,10 +163,34 @@ namespace PlayFab
...
         curl_slist* curlHttpHeaders = nullptr;
-        curlHttpHeaders = curl_slist_append(curlHttpHeaders, "Accept: application/json");
-        curlHttpHeaders = curl_slist_append(curlHttpHeaders, "Content-Type: application/json; charset=utf-8");
-        curlHttpHeaders = curl_slist_append(curlHttpHeaders, ("X-PlayFabSDK: " + PlayFabSettings::versionString).c_str());
-        curlHttpHeaders = curl_slist_append(curlHttpHeaders, "X-ReportErrorAsSuccess: true");
+
+        curlHttpHeaders = TryCurlAddHeader(reqContainer, curlHttpHeaders, "Accept: application/json");
+        if (curlHttpHeaders == NULL)
+        {
+            HandleCallback(std::move(requestContainer));
```
pgilmorepf, line 170:
```suggestion
```

source/code/source/playfab/PlayFabCurlHttpPlugin.cpp
```diff
@@ -164,10 +163,34 @@ namespace PlayFab
...
+            return;
+        }
+
+        curlHttpHeaders = TryCurlAddHeader(reqContainer, curlHttpHeaders, "Content-Type: application/json; charset=utf-8");
+        if (curlHttpHeaders == NULL)
+        {
+            HandleCallback(std::move(requestContainer));
```
pgilmorepf, line 177:
```suggestion
```

source/code/source/playfab/PlayFabCurlHttpPlugin.cpp
```diff
@@ -164,10 +163,34 @@ namespace PlayFab
...
+            return;
+        }
+
+        curlHttpHeaders = TryCurlAddHeader(reqContainer, curlHttpHeaders, ("X-PlayFabSDK: " + PlayFabSettings::versionString).c_str());
+        if (curlHttpHeaders == NULL)
+        {
+            HandleCallback(std::move(requestContainer));
```
pgilmorepf, line 184:
```suggestion
```

source/code/source/playfab/PlayFabCurlHttpPlugin.cpp
toddbellMSFT: list starts as a nullptr. This code prevents us from ever adding to the curl_slist (see line 165, it's the most commonly suggested way to be calling curl_slist_append by passing in the nullptr first https://curl.haxx.se/libcurl/c/curl_slist_append.html

This entire PR is only to check for the pointer to the list coming out. if that is a nullptr, something went wrong, and we need to end whatever we are doing.

PR #99, opened 2020-05-12
Toddbell/threaded callbacks fix

toddbellMSFT:
We may potentially throw an error synchronously, even if threaded Callbacks is set to false. There are 3 optional fixes:

A.) comment the behavior (we will synchronously call back the error if we know the http request will fail)
B.) Throw an exception (smallest possible change)
C.) Move HandleInvalidSettings after queuing the request (appropriate fix, but has the highest surface area change).

### DIFF
--- a/source/code/source/playfab/PlayFabCallRequestContainer.cpp
+++ b/source/code/source/playfab/PlayFabCallRequestContainer.cpp
@@ -77,32 +77,10 @@ namespace PlayFab
 
     bool CallRequestContainer::HandleInvalidSettings()
     {
-        bool isValid = true;
         if (m_settings->titleId.empty())
         {
-            // Option B
-            //throw new PlayFabException(PlayFabExceptionCode::TitleNotSet, "PlayFabSettings::staticSettings->titleId has not been set properly. It must not be empty.");
-
-            errorWrapper.HttpCode = 0;
-            errorWrapper.HttpStatus = "Client-side validation failure";
-            errorWrapper.ErrorCode = PlayFabErrorCode::PlayFabErrorInvalidParams;
-            errorWrapper.ErrorName = errorWrapper.HttpStatus;
-            errorWrapper.ErrorMessage = "PlayFabSettings::staticSettings->titleId has not been set properly. It must not be empty.";
-            isValid = false;
+            throw new PlayFabException(PlayFabExceptionCode::TitleNotSet, "PlayFabSettings::staticSettings->titleId has not been set properly. It must not be empty.");
         }
-
-        if (!isValid)
-        {
-            if (PlayFabSettings::globalErrorHandler != nullptr)
-            {
-                PlayFabSettings::globalErrorHandler(errorWrapper, GetCustomData());
-            }
-            if (errorCallback != nullptr)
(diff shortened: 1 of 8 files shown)


source/code/source/playfab/PlayFabSettings.cpp.ejs
```diff
@@ -5,6 +5,8 @@
 namespace PlayFab
 {
     // Control whether all callbacks are threaded or whether the user manually controls callback timing from their main-thread
+    // OPTION A-1 - If the title id is not set, the error callback will be synchronosly called within your failing api call.
+    //        B     ANY api call may synchronously throw an exception if the title id is not set
```
pgilmorepf, line 9: I like B.

source/code/source/playfab/PlayFabCallRequestContainer.cpp
```diff
@@ -80,6 +80,9 @@ namespace PlayFab
         bool isValid = true;
         if (m_settings->titleId.empty())
         {
+            // Option B
+            //throw new PlayFabException(PlayFabExceptionCode::TitleNotSet, "PlayFabSettings::staticSettings->titleId has not been set properly. It must not be empty.");
```
pgilmorepf, line 84: Option B obviously also involves deleting essentially all the rest of the code in this method.
I do like it though, because it is a more direct and instantaneous feedback, at the point of failure.
The current mechanism returns the error as a callback, which is disconnected from the point of failure.

ScottMunroMS:
agreed with @pgilmorepf 
Option B sounds like a good idea to me.
Option C sounds nice on the surface in that it simplifies things for the title so that all error handling is done in one place (the error callback), but since these errors are contract violations I think the exceptions in Option B actually sound more appropriate (despite my general distaste of exceptions)

source/code/source/playfab/PlayFabCallRequestContainer.cpp
```diff
@@ -77,29 +77,10 @@ namespace PlayFab
 
     bool CallRequestContainer::HandleInvalidSettings()
```
ScottMunroMS, line 78: should this be a void return now that it always returns true?

and perhaps the name could be updated to something that more explicitly signifies exception throwing like `ValidateSettings` or `ThrowIfSettingsInvalid`

source/code/source/playfab/PlayFabSettings.cpp.ejs
```diff
@@ -5,6 +5,7 @@
 namespace PlayFab
 {
     // Control whether all callbacks are threaded or whether the user manually controls callback timing from their main-thread
+    // Note ANY api call may synchronously throw an exception if the title id is not set, regardless of this flag
```
ScottMunroMS, line 8: "regardless of this flag" doesn't seem relevant anymore since throwing an exception doesn't violate what this flag represents

ScottMunroMS: i.e. throwing an exception isn't a callback so regardless of what this flag is set as, it should always be expected that calling an API method can throw since none of these are marked `noexcept`

PR #101, opened 2020-07-09
Adding a PlatformLoginTest to the XPlat test project

pgilmorepf:
It'll be additionally possible for individual platforms to define their own platform-relevant login tests.


pgilmorepf:
Force landing since Todd is on vacation.

PR #102, opened 2020-07-10
Adding PlatformLoginTest to Android

pgilmorepf:
Previous commit broke Android


pgilmorepf:
Force landing because Todd is unavailable.

PR #105, opened 2020-08-05
Update PlayFabQoSApi.cpp

toddbellMSFT:
Attempt to return an error result if all results are the timeout length.


source/code/source/playfab/QoS/PlayFabQoSApi.cpp
```diff
@@ -136,6 +136,22 @@ namespace PlayFab
...
+
+            bool allTimedOut = false;
+            for(auto& regionRes: result.regionResults)
+            {
+                if(regionRes.latencyMs != timeoutMs)
+                {
+                    allTimedOut = false;
+                    break;
+                }
+                allTimedOut = true;
```
pgilmorepf, line 148:
```suggestion
```

source/code/source/playfab/QoS/PlayFabQoSApi.cpp
```diff
@@ -372,6 +388,7 @@ namespace PlayFab
             }
 
             // if a ping took longer than TIMEOUT value it will be ignored (by design)
+            // THIS IS WRONG, IF ALL REGIONS ARE TIMEOUT WE SHOULD ENSURE WE ARE HANDING BACK AN ERROR
```
pgilmorepf, line 391: Is this a TODO? Is it a bug?

source/test/TestApp/PlayFabQoSTest.cpp
```diff
@@ -2,10 +2,11 @@
 
 #include "TestAppPch.h"
 
-#if defined(PLAYFAB_PLATFORM_WINDOWS) || defined(PLAYFAB_PLATFORM_XBOX)
+//#if defined(PLAYFAB_PLATFORM_WINDOWS) || defined(PLAYFAB_PLATFORM_XBOX)
```
pgilmorepf, line 5: You can't blanket remove this without thoroughly testing on every other platform.
At the moment, that's an intense process due to the way the private templates lock to specific core commits (private templates don't test latest core when you make changes to core - yet).

source/test/TestApp/PlayFabQoSTest.cpp
```diff
@@ -24,11 +25,29 @@ namespace PlayFabUnit
...
+            [&isLoggedIn](const LoginResult& loginResult, void*) {
+                PlayFabSettings::staticPlayer->clientSessionTicket = loginResult.SessionTicket;
+                isLoggedIn.exchange(true);
+            },
+            [&isLoggedIn](const PlayFabError&, void*) {
+                isLoggedIn.exchange(true);
+            },
+                &testContext);
+
+        while (!isLoggedIn) {}
```
pgilmorepf, line 46: It's not appropriate to block infinitely on login.
The http layer, or the login, or whatever, could crash somewhere, and never set this boolean.

The C++ test framework is designed to do async testing where this method kicks off the test, and then the test comes back later with a result.
Review ClientApiTests, and follow that pattern.

source/test/TestApp/PlayFabQoSTest.cpp
```diff
@@ -24,11 +25,29 @@ namespace PlayFabUnit
...
+
         QoS::PlayFabQoSApi api;
 
-        QoS::QoSResult result = api.GetQoSResult(5, 200);
+        QoS::QoSResult result = api.GetQoSResult(5);
```
pgilmorepf, line 50: Presuming that this too, still blocks infinitely until the result is gotten rather than returning from this method and waiting for the result to be set later.

source/test/TestApp/PlayFabQoSTest.cpp
```diff
@@ -24,11 +25,29 @@ namespace PlayFabUnit
...
 
         if (result.errorCode == 0)
```
pgilmorepf, line 52: And this confirms it, the test passes or fails here, which means all internals are blocking until completion rather than ever return thread and timing control back to the test framework.

source/test/TestApp/PlayFabQoSTest.cpp
```diff
@@ -41,39 +60,36 @@ namespace PlayFabUnit
     }
 #endif
 
-    void PlayFabEventTest::AddTests()
+    void PlayFabQoSTest::AddTests()
     {
-        // It is not appropriate for this test to block infinitely until the QoS result is returned
-        // Not the least of which is because it frequently blocks forever
-        // This needs to switch to the async mechanism, and allow the test to time out if QoS never completes
```
pgilmorepf, line 48: These comments are relevant, and not addressed by these changes.

source/test/TestApp/PlayFabQoSTest.h
```diff
@@ -11,18 +11,10 @@ namespace PlayFabUnit
 {
     struct TestContext;
 
-    // A wrapper around TestEvents, so we can track it back to the test that launched the event
-    class TestEvent : public PlayFab::PlayFabEvent
```
pgilmorepf, line 15: Don't remove this until you determine it's not necessary. A properly implemented async pattern on the QoS tests may require it.

source/test/TestApp/TestApp.cpp
```diff
@@ -104,6 +107,10 @@ namespace PlayFabUnit
         testRunner.Add(pfMultiUserInstanceTest);
 #endif
 
+        PlayFabQoSTest pfQosTest;
```
pgilmorepf, line 110: I believe that QoS is exclusively useful for clients. It may only be worth keeping this inside of the
```#if !defined(DISABLE_PLAYFABCLIENT_API)```
block

**Review by pgilmorepf — CHANGES_REQUESTED** [review]
The QoS tests were disabled because they were implemented wildly wrong, and they infinitely lock up the tests on multiple platforms. This causes those builds to never stop, never time out, never end, which breaks... well... everything.

Some of your changes may be useful or right, in the QoS system, but your changes to the tests have not addressed any of the documented problems with the tests (and you're removing said documentation, which will blow up all the non-windows pipelines)

File: source/test/TestApp/PlayFabQoSTest.cpp (hunk 4 above)
Comment (PushpadantK) [reply source/test/TestApp/PlayFabQoSTest.cpp:46]: +1 and also you need to yield or sleep to give a chance to another thread to do its work instead of busy waiting.

---
In reply to: [514816352](https://github.com/PlayFab/XPlatCoreTemplate/pull/105#discussion_r514816352) [](ancestors = 514816352)

File: source/test/TestApp/PlayFabQoSTest.h (hunk 8 above)
Comment (toddbellMSFT) [reply source/test/TestApp/PlayFabQoSTest.h:15]: locally, tests passed without this. 

Jenkins passed https://pf-jenkins.playfab.com/view/SDK/job/SDK_Pipeline_Filtered/3253/console without it.

File: source/test/TestApp/PlayFabQoSTest.cpp (hunk 4 above)
Comment (toddbellMSFT) [reply source/test/TestApp/PlayFabQoSTest.cpp:46]: The initial test was supposed to work in the same way as ClientApiTests, but due to the implementation of QoS, the multiplayer API call getting queued in a login call back leads to a dead lock. 

How? The thread that calls back on login tells qos to make that multiplayer call (for this tests purpose). But that is a critical point of failure for QoS since inception. QoS calls a multiplayer API to get back responses from servers. That multiplayer call gets queued on the WinHttp object's pending_requests. The issue here is, we are already on that WinHttp worker thread, completing the login call. 

Qos has always done this "block the current thread while we get callbacks to multiple playfab multiplayer calls and sort em". This API was not implemented appropriately for async, but that complete change is out of scope of this PR. 

This PR just covers getting the test passing to confirm behavior (it never passed before due to the deadlock, which probably coincides with changing over to the Instance based api's). QoS relies on the static api logic right now and that change is VERY out of scope to the particular bug we are looking at (if things time out, we should hand back a time out error code). 

We can address fixing this infinite blocking issue by re-architechting QoS to enable an instance based API version that accepts an authentication context that is required to make the multiplayer calls. That should remove the oddity of needing to wait anywhere (the qos instance should handle the callbacks for the multiplayer call, and then forward the final callback back to the customer like all of our other api's). 

---
In reply to: [514816352](https://github.com/PlayFab/XPlatCoreTemplate/pull/105#discussion_r514816352) [](ancestors = 514816352)

File: source/test/TestApp/TestApp.cpp [hunk 10]
```diff
@@ -104,6 +107,12 @@ namespace PlayFabUnit
         testRunner.Add(pfMultiUserInstanceTest);
 #endif
```
pgilmorepf, line 108: This is the end of the #if !defined(DISABLE_PLAYFABCLIENT_API) block.
(Maybe we should comment-label these endif's for clarity)

pgilmorepf, line 108: This is the end of the #if !defined(DISABLE_PLAYFABCLIENT_API) block.
(Maybe we should comment-label these endif's for clarity)

source/test/TestApp/TestApp.cpp
pgilmorepf: While limiting the platforms is good, I don't think this should run from servers either.
Please move this two lines up, into the client-only #if block, so that it's only run in client-only builds.

pgilmorepf (approved):
One very trivial additional request, shifting a code block by 2 lines to put it into the client-only tests.
Approved without need for further review after that adjustment.

PR #111, opened 2020-09-29
Moved batch so that state was retained beween DoWork calls

boczar:
Moved batch container up one step in the callstack to retain state between calls when using background threads.
Explicit ::Update() call path flushes any partial batches before exiting.


boczar:
I could also change it so that the batch is part of the class, but that would cause unexpected behavior if two threads call ::Update() at the same time.

boczar:
Added a potentially risky flush in ::Stop to avoid dropping events. Let me know if you have any concerns.

PR #114, opened 2020-10-15
Add GroupsApi test

toddbellMSFT:
Recent investigation required me to write this test, thought we may want to add it to our test suite if it is logical/correct to just create groups.

### DIFF
--- a/source/code/source/playfab/QoS/PlayFabQoSApi.cpp
+++ b/source/code/source/playfab/QoS/PlayFabQoSApi.cpp
@@ -135,7 +135,23 @@ namespace PlayFab
                 result.regionResults.push_back(RegionResult(it->first, latency, it->second.errorCode));
             }
 
+            bool allTimedOut = true;
+            for(auto& regionRes: result.regionResults)
+            {
+                if(regionRes.latencyMs != timeoutMs)
+                {
+                    allTimedOut = false;
+                    break;
+                }
+            }
+
+            if(allTimedOut)
+            {
+                result.errorCode = static_cast<int>(QoSErrorCode::Timeout);;
+            }
+
             std::sort(result.regionResults.begin(), result.regionResults.end(), [](const RegionResult& first, const RegionResult& second) -> bool {return first.latencyMs < second.latencyMs; });
+
             return result;
         }
 
@@ -289,8 +305,12 @@ namespace PlayFab
                 for (size_t i = 0; i < numThreads && pingItr < numPings; ++i)
                 {
                     // NOTE: the very first ping result might be a fake future
-                    for (size_t futuresCount = 0; futuresCount < MaxWaitForFuturesLoopCounts || asyncPingResults[i].valid(); ++futuresCount)
+                    for (size_t futuresCount = 0; futuresCount < MaxWaitForFuturesLoopCounts; ++futuresCount)
                     {
+                        if(!asyncPingResults[i].valid())
+                        {
+                            continue;
+                        }
                         future_status status = asyncPingResults[i].wait_for(threadWaitTimespan);
                         if (status == future_status::ready)
                         {
@@ -370,8 +390,6 @@ namespace PlayFab
                 accumulatedPingResults[region].latencyMs += result.latencyMs;
                 ++accumulatedPingResults[region].pingCount;
             }
-
-            // if a ping took longer than TIMEOUT value it will be ignored (by design)
         }
 
         // Ping one data center and return the address.
--- a/source/code/source/playfab/QoS/PlayFabQoSApi.cpp
+++ b/source/code/source/playfab/QoS/PlayFabQoSApi.cpp
@@ -138,7 +138,7 @@ namespace PlayFab
             bool allTimedOut = true;
             for(auto& regionRes: result.regionResults)
             {
-                if(regionRes.latencyMs != timeoutMs)
+                if(regionRes.latencyMs != static_cast<int>(timeoutMs))
                 {
                     allTimedOut = false;
                     break;
--- a/source/test/TestApp/PlayFabApiTest.cpp
+++ b/source/test/TestApp/PlayFabApiTest.cpp
@@ -22,8 +22,6 @@ using namespace ClientModels;
 using namespace AuthenticationModels;
 using namespace DataModels;
 
-
-
 namespace PlayFabUnit
 {
     const std::string PlayFabApiTest::TEST_DATA_KEY = "testCounter";
@@ -733,41 +731,41 @@ namespace PlayFabUnit
         }
     }
 
-     // GROUPS API
-    void PlayFabApiTest::GroupsApiTest(TestContext& testContext)
-    {
-        auto settings = std::make_shared<PlayFabApiSettings>();
+    // // GROUPS API
+    // void PlayFabApiTest::GroupsApiTest(TestContext& testContext)
+    // {
+    //     auto settings = std::make_shared<PlayFabApiSettings>();
 
-        auto req = PlayFab::ClientModels::LoginWithCustomIDRequest();
-        req.CustomId = PlayFabSettings::buildIdentifier;
+    //     auto req = PlayFab::ClientModels::LoginWithCustomIDRequest();
+    //     req.CustomId = PlayFabSettings::buildIdentifier;
 
-        clientApi->LoginWithCustomID(req, Callback(&PlayFabApiTest::GroupsTestLoginCallback), Callback(&PlayFabApiTest::GroupsTestLoginFailedCallback), &testContext);
-    }
+    //     clientApi->LoginWithCustomID(req, Callback(&PlayFabApiTest::GroupsTestLoginCallback), Callback(&PlayFabApiTest::GroupsTestLoginFailedCallback), &testContext);
+    // }
 
-    void PlayFabApiTest::GroupsTestLoginCallback(const LoginResult& result, void* customData)
-    {
-        auto groupApi = std::make_shared<PlayFab::PlayFabGroupsInstanceAPI>(result.authenticationContext);
-        auto req = PlayFab::GroupsModels::CreateGroupRequest();
-        req.CustomTags = std::map<std::string, std::string>();
+    // void PlayFabApiTest::GroupsTestLoginCallback(const LoginResult& result, void* customData)
+    // {
+    //     auto groupApi = std::make_shared<PlayFab::PlayFabGroupsInstanceAPI>(result.authenticationContext);
+    //     auto req = PlayFab::GroupsModels::CreateGroupRequest();
+    //     req.CustomTags = std::map<std::string, std::string>();
 
-        // TODO Bug 29786037: this map is required to be filled to not get a 500 error with the CreateGroup api call
-        req.CustomTags.insert(std::pair<std::string, std::string>("One", "Two"));
+    //     // TODO Bug 29786037: this map is required to be filled to not get a 500 error with the CreateGroup api call
+    //     req.CustomTags.insert(std::pair<std::string, std::string>("One", "Two"));
 
-        req.GroupName = PlayFabUnit::GenerateUuidV4();
-        groupApi->CreateGroup(req, Callback(&PlayFabApiTest::GroupsTestGroupCallback), Callback(&PlayFabApiTest::GroupsTestLoginFailedCallback), customData);
-    }
+    //     req.GroupName = PlayFabUnit::GenerateUuidV4();
+    //     groupApi->CreateGroup(req, Callback(&PlayFabApiTest::GroupsTestGroupCallback), Callback(&PlayFabApiTest::GroupsTestLoginFailedCallback), customData);
+    // }
 
-    void PlayFabApiTest::GroupsTestGroupCallback(const PlayFab::GroupsModels::CreateGroupResponse& response, void* customData)
-    {
-        TestContext* testContext = static_cast<TestContext*>(customData);
-        testContext->Pass("CreateGroups succeeded and made the new group: " + response.GroupName);
-    }
+    // void PlayFabApiTest::GroupsTestGroupCallback(const PlayFab::GroupsModels::CreateGroupResponse& response, void* customData)
+    // {
+    //     TestContext* testContext = static_cast<TestContext*>(customData);
+    //     testContext->Pass("CreateGroups succeeded and made the new group: " + response.GroupName);
+    // }
 
-    void PlayFabApiTest::GroupsTestLoginFailedCallback(const PlayFabError& error, void* customData)
-    {
-        TestContext* testContext = static_cast<TestContext*>(customData);
-        testContext->Fail("Expected Group Login to succeed. Got error " + error.ErrorMessage);
-    }
+    // void PlayFabApiTest::GroupsTestLoginFailedCallback(const PlayFabError& error, void* customData)
+    // {
+    //     TestContext* testContext = static_cast<TestContext*>(customData);
+    //     testContext->Fail("Expected Group Login to succeed. Got error " + error.ErrorMessage);
+    // }
 
     //
     //
@@ -794,7 +792,7 @@ namespace PlayFabUnit
         AddTest("WriteEvent", &PlayFabApiTest::WriteEvent);
         AddTest("GetEntityToken", &PlayFabApiTest::GetEntityToken);
         AddTest("ObjectApi", &PlayFabApiTest::ObjectApi);
-        AddTest("CreateGroupApi", &PlayFabApiTest::GroupsApiTest);
+        //AddTest("CreateGroupApi", &PlayFabApiTest::GroupsApiTest);
     }
 
     void PlayFabApiTest::ClassSetUp()
--- a/source/test/TestApp/PlayFabApiTest.cpp
+++ b/source/test/TestApp/PlayFabApiTest.cpp
@@ -731,42 +731,6 @@ namespace PlayFabUnit
         }
     }
 
-    // // GROUPS API
-    // void PlayFabApiTest::GroupsApiTest(TestContext& testContext)
-    // {
-    //     auto settings = std::make_shared<PlayFabApiSettings>();
-
-    //     auto req = PlayFab::ClientModels::LoginWithCustomIDRequest();
-    //     req.CustomId = PlayFabSettings::buildIdentifier;
-
-    //     clientApi->LoginWithCustomID(req, Callback(&PlayFabApiTest::GroupsTestLoginCallback), Callback(&PlayFabApiTest::GroupsTestLoginFailedCallback), &testContext);
-    // }
-
-    // void PlayFabApiTest::GroupsTestLoginCallback(const LoginResult& result, void* customData)
-    // {
-    //     auto groupApi = std::make_shared<PlayFab::PlayFabGroupsInstanceAPI>(result.authenticationContext);
-    //     auto req = PlayFab::GroupsModels::CreateGroupRequest();
-    //     req.CustomTags = std::map<std::string, std::string>();
-
-    //     // TODO Bug 29786037: this map is required to be filled to not get a 500 error with the CreateGroup api call
-    //     req.CustomTags.insert(std::pair<std::string, std::string>("One", "Two"));
-
-    //     req.GroupName = PlayFabUnit::GenerateUuidV4();
-    //     groupApi->CreateGroup(req, Callback(&PlayFabApiTest::GroupsTestGroupCallback), Callback(&PlayFabApiTest::GroupsTestLoginFailedCallback), customData);
-    // }
-
-    // void PlayFabApiTest::GroupsTestGroupCallback(const PlayFab::GroupsModels::CreateGroupResponse& response, void* customData)
-    // {
-    //     TestContext* testContext = static_cast<TestContext*>(customData);
-    //     testContext->Pass("CreateGroups succeeded and made the new group: " + response.GroupName);
-    // }
-
-    // void PlayFabApiTest::GroupsTestLoginFailedCallback(const PlayFabError& error, void* customData)
-    // {
-    //     TestContext* testContext = static_cast<TestContext*>(customData);
-    //     testContext->Fail("Expected Group Login to succeed. Got error " + error.ErrorMessage);
-    // }
-
     //
     //
     // Add test calls to this method, after implementation
@@ -792,7 +756,6 @@ namespace PlayFabUnit
         AddTest("WriteEvent", &PlayFabApiTest::WriteEvent);
         AddTest("GetEntityToken", &PlayFabApiTest::GetEntityToken);
         AddTest("ObjectApi", &PlayFabApiTest::ObjectApi);
-        //AddTest("CreateGroupApi", &PlayFabApiTest::GroupsApiTest);
     }
 
     void PlayFabApiTest::ClassSetUp()
--- a/source/test/TestApp/PlayFabApiTest.cpp
+++ b/source/test/TestApp/PlayFabApiTest.cpp
@@ -730,6 +730,38 @@ namespace PlayFabUnit
         }
     }
 
+    // GROUP API
+    void PlayFabApiTest::GroupEmptyTest(TestContext& testContext)
+    {
+        // TODO need to add Group API (does it need to be imported first?)
+        // Login (need to define entity ID and title_player_account) and group too?
+        //   call CreateGroup 
+        //   Ensure CustomTags is in the header and that it is empty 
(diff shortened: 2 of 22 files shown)


source/test/TestApp/PlayFabApiTest.cpp
```diff
@@ -16,11 +16,48 @@
...
 using namespace AuthenticationModels;
 using namespace DataModels;
 
+namespace uuid {
+    static std::random_device              rd;
+    static std::mt19937                    gen(rd());
+    static std::uniform_int_distribution<> dis(0, 15);
+    static std::uniform_int_distribution<> dis2(8, 11);
+
+    std::string generate_uuid_v4() {
```
pgilmorepf, line 33: Generating a random guid would be a great general utility, not just a random method in a test.

source/test/TestApp/PlayFabApiTest.cpp
```diff
@@ -730,6 +767,43 @@ namespace PlayFabUnit
         }
     }
 
+     // GROUPS API
+    void PlayFabApiTest::GroupsApiTest(TestContext& testContext)
```
pgilmorepf, line 771: This new set of tests should be a whole new test file: PlayFabGroupsTest.cpp/.h

I would have many more comments and changes to suggest if we don't do that and leave it here. But I think if we separate this to a separate file & test suite, we'd negate those other comments, so I'll just make this suggestion alone.

source/test/TestApp/PlayFabApiTest.cpp
```diff
@@ -730,6 +767,43 @@ namespace PlayFabUnit
...
+    {
+        auto settings = std::make_shared<PlayFabApiSettings>();
+        //clientApi = std::make_shared<PlayFab::PlayFabClientInstanceAPI>(settings);
```
pgilmorepf, line 774: Delete unused code comments.
Also, this is definitely something that should only exist in a setup step, not in the middle of a test.

pgilmorepf (changes requested):
Main comment is just to create a new suite and put the new tests there.

source/test/TestApp/PlayFabGroupsTest.cpp
```diff
@@ -0,0 +1,85 @@
...
+    void PlayFabGroupsTest::AddTests()
+    {
+        AddTest("CreateGroupApi", &PlayFabGroupsTest::GroupsApiTest);
+    }
+
+    void PlayFabGroupsTest::ClassSetUp()
+    {
+        clientApi = std::make_shared<PlayFabClientInstanceAPI>(PlayFabSettings::staticPlayer);
+        PlayFabSettings::staticSettings->titleId = testTitleData.titleId;
+        USER_EMAIL = testTitleData.userEmail;
```
pgilmorepf, line 58: This test does not need USER_EMAIL

source/test/TestApp/PlayFabGroupsTest.cpp
```diff
@@ -0,0 +1,85 @@
...
+
+        // Verify all the inputs won't cause crashes in the tests
+        TITLE_INFO_SET = !PlayFabSettings::staticSettings->titleId.empty() && !USER_EMAIL.empty();
```
pgilmorepf, line 61: This test does not need USER_EMAIL

source/test/TestApp/PlayFabGroupsTest.h
```diff
@@ -0,0 +1,46 @@
...
+#include "playfab/PlayFabGroupsDataModels.h"
+
+namespace PlayFabUnit
+{
+    class PlayFabGroupsTest : public PlayFabApiTestCase
+    {
+    private:
+        // Fixed values provided from testInputs
+        bool TITLE_INFO_SET;
+        std::string USER_EMAIL;
```
pgilmorepf, line 19: This test does not need USER_EMAIL

source/test/TestApp/PlayFabGroupsTest.cpp
```diff
@@ -0,0 +1,85 @@
...
+    void PlayFabGroupsTest::GroupsTestGroupCallback(const GroupsModels::CreateGroupResponse&, void* customData)
+    {
+        TestContext* testContext = static_cast<TestContext*>(customData);
+        testContext->Pass();
+    }
+
+    void PlayFabGroupsTest::GroupsTestLoginFailedCallback(const PlayFabError& error, void* customData)
+    {
+        TestContext* testContext = static_cast<TestContext*>(customData);
+        testContext->Fail("Expected Group Login to succeed. Got error " + error.ErrorMessage);
```
pgilmorepf, line 46: error.GenerateErrorReport()

source/test/TestApp/PlayFabGroupsTest.cpp
```diff
@@ -0,0 +1,85 @@
...
+        groupApi->CreateGroup(req, Callback(&PlayFabGroupsTest::GroupsTestGroupCallback), Callback(&PlayFabGroupsTest::GroupsTestLoginFailedCallback), customData);
+    }
+
+    void PlayFabGroupsTest::GroupsTestGroupCallback(const GroupsModels::CreateGroupResponse&, void* customData)
+    {
+        TestContext* testContext = static_cast<TestContext*>(customData);
+        testContext->Pass();
+    }
+
+    void PlayFabGroupsTest::GroupsTestLoginFailedCallback(const PlayFabError& error, void* customData)
```
pgilmorepf, line 43: Rename:
If this is meant to be a shared callback (as it is used), then it should be renamed to be described as a shared callback, and the error message 3 lines down should be restated to not indicate an explicit login failure.
If this is NOT meant to be a shared callback (as it is named & strings within indicate), then you need to write a separate method as the fail callback for the CreateGroup call.

source/test/TestApp/PlayFabGroupsTest.cpp
```diff
@@ -0,0 +1,85 @@
...
+    {
+        auto groupApi = std::make_shared<PlayFabGroupsInstanceAPI>(result.authenticationContext);
+        auto req = GroupsModels::CreateGroupRequest();
+        req.CustomTags = std::map<std::string, std::string>();
+
+        // TODO Bug 29786037: this map is required to be filled to not get a 500 error with the CreateGroup api call
+        req.CustomTags.insert(std::pair<std::string, std::string>("One", "Two"));
+
+        req.GroupName = GenerateUuidV4();
+        groupApi->CreateGroup(req, Callback(&PlayFabGroupsTest::GroupsTestGroupCallback), Callback(&PlayFabGroupsTest::GroupsTestLoginFailedCallback), customData);
```
pgilmorepf, line 34: This reuses the GroupsTestLoginFailedCallback callback, which by name, and by text within that method doesn't reflect it as generic reusable. See the mirror comment below on that method.

source/test/TestApp/PlayFabGroupsTest.cpp
```diff
@@ -0,0 +1,85 @@
...
+
+    void PlayFabGroupsTest::GroupsTestLoginFailedCallback(const PlayFabError& error, void* customData)
+    {
+        TestContext* testContext = static_cast<TestContext*>(customData);
+        testContext->Fail("Expected Group Login to succeed. Got error " + error.ErrorMessage);
+    }
+
+    void PlayFabGroupsTest::AddTests()
+    {
+        AddTest("CreateGroupApi", &PlayFabGroupsTest::GroupsApiTest);
```
pgilmorepf, line 51: I think a better plan here would be to separate the login step into it's own test, and put two tests here.
That gives each call an independent test-timeout, and is a little safer in terms of validating each call independently.
It's true that the "login" test isn't a test of groups, but as we add tests, we don't want to duplicate login logic.

For individual suites we've been very comfortable with the in-suite ordering being strict and reliable.

source/test/TestApp/TestUtils.cpp
```diff
@@ -3,8 +3,44 @@
 #include "TestAppPch.h"
 #include "TestUtils.h"
 
+#include <random>
+#include <sstream>
+
 namespace PlayFabUnit
 {
+    
+    std::string GenerateUuidV4() {
```
pgilmorepf, line 12: I think that "TestUtils" is a big improvement, but I'll note that the main SDK and thus most customers still won't have direct access to this.
I think I would have been happier with a generic PlayFabUtil.GenerateUuidV4.

source/test/TestApp/PlayFabGroupsTest.cpp
```diff
@@ -0,0 +1,85 @@
...
+    void PlayFabGroupsTest::GroupsTestLoginCallback(const ClientModels::LoginResult& result, void* customData)
+    {
+        auto groupApi = std::make_shared<PlayFabGroupsInstanceAPI>(result.authenticationContext);
+        auto req = GroupsModels::CreateGroupRequest();
+        req.CustomTags = std::map<std::string, std::string>();
+
+        // TODO Bug 29786037: this map is required to be filled to not get a 500 error with the CreateGroup api call
+        req.CustomTags.insert(std::pair<std::string, std::string>("One", "Two"));
+
+        req.GroupName = GenerateUuidV4();
```
pgilmorepf, line 33: Because this is the only usage of GenerateUuidV4, and as I state down there, it's not "sufficiently random" to be a GUID, I think we should re-evaluate either the name of this method, or its implementation.

source/test/TestApp/TestUtils.cpp
```diff
@@ -3,8 +3,44 @@
...
+        std::random_device              rd;
+        std::mt19937                    gen(rd());
```
pgilmorepf, line 14: std::mt19937 defines a "source" with 32 bits of randomness.
https://en.cppreference.com/w/cpp/numeric/random/mersenne_twister_engine

source/test/TestApp/TestUtils.cpp
```diff
@@ -3,8 +3,44 @@
...
+        std::uniform_int_distribution<> dis(0, 15);
+        std::uniform_int_distribution<> dis2(8, 11);
+
+        std::stringstream ss;
+        int i;
+        ss << std::hex;
+        for (i = 0; i < 8; i++) {
+            ss << dis(gen);
```
pgilmorepf, line 22: By calling the same randomizer 4 times in sequence, you're not actually generating guid's as randomly as guids are supposed to be.
Guids have 128 bits of randomness, but in your case, your method only has 32.
Repeatedly calling dis() sequentially won't be 128 bits of randomness, but only 32, because only the first number is "random", since it was seeded by the time. Every number after the first one is predicatable, and based on only the first number.

Meaning that this GUID generator only has 4E9 (4 billion) possible outputs rather than the intended 3E38 (300 undecillion or 300 billion billion billion billion) outputs.

In short, it generates a number that is in guid format, but it's not actually a reliable guid in terms of being super unique. In fact, it's extremely likely to be not unique at all, and if our customers rely on these values being unique, they'll very quickly be disappointed.

pgilmorepf (changes requested):
I think that overall the intent here is very good, but the GUID generator fails to meet the implied expectations.
EITHER: We make a real GenerateUuidV4 method, which is truly unique to 128 bits, or we rename/move the existing method to simply say "GenerateRandomGroupName".

This implementation is more than fine for a random test group name, but insufficient for true GUIDs, if we want to make them visible to customers.

I apologize that this is contradictory to my earlier instructions on the GUID method, but I didn't look at the implementation closely enough last time.

source/test/TestApp/TestUtils.cpp
pgilmorepf: The simplest improvement is to convert to the 64 bit mt19937_64 option.
Upgrading that to 128 bits is awkward though, as it requires creating two of these, in a way where one isn't biased against the other. I think that it requires putting a sleep between the two, such as a 1ms sleep (which will be randomized by how long the operating system actually chooses to sleep, likely >1ms). That MAY be "sufficient" to get "more than" 64 bits of randomness, but I'm not sure it'll be the full 128 bits. You might have to do some research on how the seed is generated in order to get randomness at the full promised 128 bits.

toddbellMSFT: I pulled this info from stack overflow. I'm leaving a note about the source and moving on. This is sufficient for the test we are adding.

source/test/TestApp/TestUtils.cpp
toddbellMSFT: renamed to not call it a guid.

source/test/TestApp/PlayFabGroupsTest.cpp
toddbellMSFT: Other login tests in previous test suites have been on a per test basis. 

If we want to combine logins for additional tests in the future we can do it then.

source/test/TestApp/PlayFabGroupsTest.cpp
```diff
@@ -0,0 +1,84 @@
...
+
+using namespace PlayFab;
+
+namespace PlayFabUnit
+{
+
+    // GROUPS API
+    void PlayFabGroupsTest::GroupsApiTest(TestContext& testContext)
+    {
+        auto settings = std::make_shared<PlayFabApiSettings>();
```
pgilmorepf, line 16: Unused.

source/test/TestApp/PlayFabGroupsTest.cpp
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    void PlayFabGroupsTest::AddTests()
+    {
+        AddTest("CreateGroupApi", &PlayFabGroupsTest::GroupsApiTest);
+    }
+
+    void PlayFabGroupsTest::ClassSetUp()
+    {
+        clientApi = std::make_shared<PlayFabClientInstanceAPI>(PlayFabSettings::staticPlayer);
```
pgilmorepf, line 56: Looks like it may have been intended to use the static settings from the test above, but failed to move it here.

pgilmorepf: Likewise, it's better test isolation if this is initialized with a local settings and context, rather than using staticPlayer and staticContext, which can have interference from other tests.

pgilmorepf (changes requested):
One more small adjustment.
Let's finish the test isolation by using local settings and context, rather than staticSettings and staticPlayer.

source/code/source/playfab/QoS/PlayFabQoSApi.cpp
```diff
@@ -138,7 +138,7 @@ namespace PlayFab
             bool allTimedOut = true;
             for(auto& regionRes: result.regionResults)
             {
-                if(regionRes.latencyMs != timeoutMs)
+                if(regionRes.latencyMs != static_cast<int>(timeoutMs))
                 {
```
MichaelMackendMS, line 142: what is this block of code checking exactly?  is there some special value for latencyMs that timeoutMs needs to match in order to be considered timed out?

source/test/TestApp/PlayFabGroupsTest.cpp
```diff
@@ -0,0 +1,87 @@
...
+        req.CustomTags = std::map<std::string, std::string>();
+
+        // TODO Bug 29786037: this map is required to be filled to not get a 500 error with the CreateGroup api call
+        req.CustomTags.insert(std::pair<std::string, std::string>("One", "Two"));
+
+        req.GroupName = GenerateRandomString();
+        groupApi->CreateGroup(req, Callback(&PlayFabGroupsTest::GroupsTestGroupCallback), Callback(&PlayFabGroupsTest::GroupsTestSharedFailureCallback), customData);
+    }
+
+    void PlayFabGroupsTest::GroupsTestGroupCallback(const GroupsModels::CreateGroupResponse&, void* customData)
```
MichaelMackendMS, line 35: This would appear to be strictly a success callback.  Can we rename this to more accurately reflect it's purpose?

source/test/TestApp/PlayFabGroupsTest.cpp
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    void PlayFabGroupsTest::ClassSetUp()
+    {
+        groupsTestSettings = std::make_shared<PlayFabApiSettings>();
+        groupsTestSettings->titleId = testTitleData.titleId;
+
+        clientApi = std::make_shared<PlayFabClientInstanceAPI>(groupsTestSettings);
+
+        // Bug 45669 - this is required for groups tests to pass.
```
MichaelMackendMS, line 59: Why is this a bug?

source/test/TestApp/TestUtils.cpp
```diff
@@ -3,8 +3,45 @@
 #include "TestAppPch.h"
 #include "TestUtils.h"
 
+#include <random>
```
MichaelMackendMS, line 6: Seems like this is also using a pch.  Any reason not to put these includes into the pch?

MichaelMackendMS (changes requested):

source/code/source/playfab/QoS/PlayFabQoSApi.cpp
toddbellMSFT: If we time out, regionRes.latencyMs is expected to have hit the timeout that users have set elsewhere (Passed in as a parameter). Rather than change the signature elsewhere (which I think may be a public facing API), I silenced a warning by adding this static cast.

source/test/TestApp/PlayFabGroupsTest.cpp
toddbellMSFT: without that map, there is an issue on the server that returns a 500 error (not an "invalid parameter", it is a separate issue that this test is supposed to help isolate for future investigation.

toddbellMSFT: We should be able to run these tests with local values, but attempting to do so ran into odd timeout issues that I wanted to investigate at a later time. It seems like a threading issue since debugging will pass the test, but running without stepping through times the test out. It looks very similar to an issue I was seeing with QoS, but it is out of scope to this particular PR.

MichaelMackendMS (approved):
thx for addressing the things!

source/test/TestApp/PlayFabGroupsTest.cpp
pgilmorepf: Actually, That is my point. Every other suite has login as a separate test & step. This is the only suite that would combine them, and I don't think it's a good idea. And, since this is a new suite, I don't think we should create it from scratch the wrong way, regardless.

source/test/TestApp/PlayFabGroupsTest.cpp
```diff
@@ -0,0 +1,87 @@
...
+    {
+        AddTest("CreateGroupApi", &PlayFabGroupsTest::GroupsApiTest);
+    }
+
+    void PlayFabGroupsTest::ClassSetUp()
+    {
+        groupsTestSettings = std::make_shared<PlayFabApiSettings>();
+        groupsTestSettings->titleId = testTitleData.titleId;
+
+        clientApi = std::make_shared<PlayFabClientInstanceAPI>(groupsTestSettings);
```
pgilmorepf, line 57: I could go either way but:
This constructor makes the clientApi utilize the staticPlayer, rather than a context defined in this test.
2 lines up, create an additional PlayFabAuthenticationContext, and pass it as a second parameter here.

pgilmorepf: Additionally, the groupApi should be defined here, and not as a 1-off in the test above.

source/test/TestApp/PlayFabGroupsTest.cpp
```diff
@@ -0,0 +1,87 @@
...
+    {
+        auto req = ClientModels::LoginWithCustomIDRequest();
+        req.CustomId = PlayFabSettings::buildIdentifier;
+
+        clientApi->LoginWithCustomID(req, Callback(&PlayFabGroupsTest::GroupsTestLoginCallback), Callback(&PlayFabGroupsTest::GroupsTestSharedFailureCallback), &testContext);
+    }
+
+    void PlayFabGroupsTest::GroupsTestLoginCallback(const ClientModels::LoginResult& result, void* customData)
+    {
+        auto groupApi = std::make_shared<PlayFabGroupsInstanceAPI>(result.authenticationContext);
```
pgilmorepf, line 24: Move this into the ClassSetup, move the groupApi variable into the test class, and construct them both with shared settings and context.
I think that's why you were having such issues with the titleID and whatnot.
This one is being created with staticSettings, and clientApi is being created with a custom context, so they're not linked to the same information.

You're constructing the API instances completely differently, and that's the source of your data desync.
Your solution of result.authenticationContext HELPS, but it fails to share settings.
It's better to just create them together in ClassSetup.

source/test/TestApp/PlayFabGroupsTest.h
```diff
@@ -0,0 +1,50 @@
...
+
+        void GroupsTestLoginCallback(const PlayFab::ClientModels::LoginResult& result, void* customData);
+        void GroupsTestSuccessCallback(const PlayFab::GroupsModels::CreateGroupResponse& response, void* customData);
+        void GroupsTestSharedFailureCallback(const PlayFab::PlayFabError& error, void* customData);
+
+    protected:
+        void AddTests() override;
+
+    public:
+        std::shared_ptr<PlayFab::PlayFabClientInstanceAPI> clientApi;
```
pgilmorepf, line 41: add groupApi definition here.
```suggestion
        std::shared_ptr<PlayFab::PlayFabClientInstanceAPI> clientApi;
        std::shared_ptr<PlayFab::PlayFabGroupInstanceAPI> groupApi;
```

pgilmorepf (changes requested):
I figured out why you were having data inconsistency problems with your titleID.
You're not using the same settings between your login and your create group.

source/test/TestApp/TestUtils.cpp
```diff
@@ -1,10 +1,43 @@
...
 #include "TestAppPch.h"
-#include "TestUtils.h"
 
 namespace PlayFabUnit
 {
+    // source: https://stackoverflow.com/questions/24365331/how-can-i-generate-uuid-in-c-without-using-boost-library
+    // note this is not sufficiently random enough to be guid's, but random strings is the original need for our tests.
+    std::string GenerateRandomString() {
+        std::random_device              rd;
+        std::mt19937                    gen(rd());
```
toddbellMSFT, line 11: _x64

source/test/TestApp/PlayFabGroupsTest.h
```diff
@@ -0,0 +1,53 @@
...
+    protected:
+        std::shared_ptr<PlayFab::PlayFabClientInstanceAPI> clientApi;
+        std::shared_ptr<PlayFab::PlayFabGroupsInstanceAPI> groupsApi;
+
+        PlayFab::GroupsModels::EntityKey createdGroupKey;
+
+        void AddTests() override;
+
+    public:
+       void ClassSetUp() override;
```
pgilmorepf, line 46: tabbing inconsistency. Press Ctrl+k+d to resolve all spacing issues with the whole file.

source/test/TestApp/PlayFabGroupsTest.h
```diff
@@ -0,0 +1,53 @@
...
+            return std::bind(func, this, std::placeholders::_1, std::placeholders::_2);
+        }
+
+    protected:
+        std::shared_ptr<PlayFab::PlayFabClientInstanceAPI> clientApi;
+        std::shared_ptr<PlayFab::PlayFabGroupsInstanceAPI> groupsApi;
+
+        PlayFab::GroupsModels::EntityKey createdGroupKey;
+
+        void AddTests() override;
```
pgilmorepf, line 43: *headscratch* Is AddTests a protected function? I thought it would be public...

source/test/TestApp/PlayFabGroupsTest.cpp
```diff
@@ -0,0 +1,109 @@
...
+        testContext->Pass();
+    }
+
+    void PlayFabGroupsTest::GroupsRemoveTest(TestContext& testContext)
+    {
+        auto req = GroupsModels::DeleteGroupRequest();
+        req.CustomTags = std::map<std::string, std::string>();
+
+        // TODO Bug 29786037: this map is required to be filled to not get a 500 error with the CreateGroup api call
+        req.CustomTags.insert(std::pair<std::string, std::string>("One", "Two"));
```
pgilmorepf, line 53: Wait, this is an issue with delete too?

pgilmorepf (approved):
Trivial notes. Maybe something actionable, especially the spacing issue, but otherwise ready to go.

Arnou2inf (approved):

toddbellMSFT:
This was deemed out of scope. Closing.

PR #117, opened 2020-12-01
added a debug ItemGroup and ItemDefinitionGroup to targets

toddbellMSFT:
also update the nuspec to also reference the appropriate debug vs release lib's

### DIFF
--- a/source/com.playfab.xplatcppsdk.v141.nuspec.ejs
+++ b/source/com.playfab.xplatcppsdk.v141.nuspec.ejs
@@ -16,12 +16,9 @@
     </metadata>
     <files>
         <file src=".\com.playfab.xplatcppsdk.<%- vsVer %>.targets" target="build\native\com.playfab.xplatcppsdk.<%- vsVer %>.targets" />
-        <file src=".\external\jsoncpp-build\x64\Debug\lib_json.lib" target="build\native\lib\x64\Debug\lib_json.lib" />
-        <file src=".\build\Windows\x64\Debug\XPlatCppWindows\XPlatCppWindows.pdb" target="build\native\lib\x64\Debug\XPlatCppWindows.pdb" />
-        <file src=".\build\Windows\x64\Debug\XPlatCppWindows\XPlatCppWindows.lib" target="build\native\lib\x64\Debug\XPlatCppWindows.lib" />
-        <file src=".\external\jsoncpp-build\x64\Release\lib_json.lib" target="build\native\lib\x64\Release\lib_json.lib" />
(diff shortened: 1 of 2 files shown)


source/com.playfab.xplatcppsdk.v141.nuspec.ejs
```diff
@@ -16,12 +16,9 @@
...
-        <file src=".\external\jsoncpp-build\x64\Debug\lib_json.lib" target="build\native\lib\x64\Debug\lib_json.lib" />
-        <file src=".\build\Windows\x64\Debug\XPlatCppWindows\XPlatCppWindows.pdb" target="build\native\lib\x64\Debug\XPlatCppWindows.pdb" />
-        <file src=".\build\Windows\x64\Debug\XPlatCppWindows\XPlatCppWindows.lib" target="build\native\lib\x64\Debug\XPlatCppWindows.lib" />
-        <file src=".\external\jsoncpp-build\x64\Release\lib_json.lib" target="build\native\lib\x64\Release\lib_json.lib" />
-        <file src=".\build\Windows\x64\Release\XPlatCppWindows\XPlatCppWindows.pdb" target="build\native\lib\x64\Release\XPlatCppWindows.pdb" />
-        <file src=".\build\Windows\x64\Release\XPlatCppWindows\XPlatCppWindows.lib" target="build\native\lib\x64\Release\XPlatCppWindows.lib" />
+        <file src=".\external\jsoncpp-build\$configuration$\lib_json.lib" target="build\native\lib\$configuration$\lib_json.lib" />
+        <file src=".\build\Windows\$configuration$\XPlatCppWindows\XPlatCppWindows.pdb" target="build\native\lib\$configuration$\XPlatCppWindows.pdb" />
```
pgilmorepf, line 20: Release should not contain a pdb?

pgilmorepf (approved):
Just a question. Approved either way after you answer it.

source/com.playfab.xplatcppsdk.v141.nuspec.ejs
toddbellMSFT: In theory, Release mode won't be step debugged due to possible optimizations that makes the expierience a pain (so swap to debug). But from this stack overflow post: https://stackoverflow.com/questions/5457095/release-generating-pdb-files-why the first comment has 2 really good points that I have ran into in the far past:
1.) when a crash report comes in from the wild, it can have information attached to debug it. 
2.) Customers can debug it themselves without us (sometimes they only pull off a release build, kinda like us)

And the accepted answer has an interesting note about Profiling should always be in Release, which the original pdb will help.

So unless we have concern for nuget package size or some possible security issue in distributing the release PDB, I think a Release pdb can still be beneficial to our customers.

PR #119, opened 2021-02-11
Adding toggle for extra tests

toddbellMSFT:


pgilmorepf (approved):
I would call this a "novel short term solution". I certainly wouldn't have thought of it.
The reason I say short term is because ideally we wouldn't be skipping tests on specific platforms.

Some questions:
Does "PLAYFAB_PLATFORM_WIN_GDK" exist? Could we properly detect it and create it?
Can we do the same for other tented GDK platforms?

Once we do that, we could use the platform define filters that we've already committed to using, so that we don't have two solutions.

I approve this for now, with the idea that we consolidate on the existing solution as a uniform solution later.

make.js
```diff
@@ -24,7 +24,8 @@ exports.makeCombinedAPI = function (apis, sourceDir, apiOutputDir) {
         vsVer: "v141", // As C++ versions change, we may need to update this
         vsYear: "2017", // As VS versions change, we may need to update this
         getVerticalNameDefault: getVerticalNameDefault,
-        winSdkVersion: "10.0.17763.0" // Which version of the Windows SDK (A VS installation option) to use
+        winSdkVersion: "10.0.17763.0", // Which version of the Windows SDK (A VS installation option) to use
+        ifNeedsExtraTests: "true"
```
MichaelMackendMS, line 28: The other parameters here have commenting. Can you add comments to explain the purpose of this value?

MichaelMackendMS (changes requested):
I think there needs to be more clarity surrounding what / when a test is considered "extra".

make.js
toddbellMSFT: removing this param in favor of doing a precompiler check for gdk

PR #120, opened 2021-02-12
Update TestApp.cpp

toddbellMSFT:
Intellisense was throwing this fix off.

### DIFF
--- a/source/test/TestApp/TestApp.cpp
+++ b/source/test/TestApp/TestApp.cpp
@@ -81,7 +81,7 @@ namespace PlayFabUnit
         // Set this up for use when the tests finish
         this->clientApi = std::make_shared<PlayFab::PlayFabClientInstanceAPI>(PlayFab::PlayFabSettings::staticPlayer);
 
-#if !defined(PLAYFAB_PLATFORM_IOS) // IOS proj doesn't include PlatformLoginTest files yet
+#if !defined(PLAYFAB_PLATFORM_GDK) && !defined(PLAYFAB_PLATFORM_IOS) // IOS proj doesn't include PlatformLoginTest files yet
         PlatformLoginTest loginTest;
         loginTest.SetTitleInfo(testTitleData);
         testRunner.Add(loginTest);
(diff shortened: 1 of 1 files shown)


source/test/TestApp/TestApp.cpp
```diff
@@ -81,7 +81,7 @@ namespace PlayFabUnit
         // Set this up for use when the tests finish
         this->clientApi = std::make_shared<PlayFab::PlayFabClientInstanceAPI>(PlayFab::PlayFabSettings::staticPlayer);
 
-#if !defined(PLAYFAB_PLATFORM_IOS) // IOS proj doesn't include PlatformLoginTest files yet
+#if !defined(PLAYFAB_PLATFORM_GDK) && !defined(PLAYFAB_PLATFORM_IOS) // IOS proj doesn't include PlatformLoginTest files yet
```
PushpadantK, line 84: >IOS [](start = 72, length = 3)

update this comment, is there another bug tracking fix for GDK?

source/test/TestApp/TestApp.cpp
```diff
@@ -92,7 +92,7 @@ namespace PlayFabUnit
         pfApiTest.SetTitleInfo(testTitleData);
         testRunner.Add(pfApiTest);
 
-#if !defined(PLAYFAB_PLATFORM_GDK) && false // These tests are still too unstable, and despite passing nearly 100% in debug, still fail 100% in release
```
PushpadantK, line 95: >false [](start = 38, length = 5)

nit : this condition will always resolve to false anyways.

PushpadantK (approved):
:shipit:

source/test/TestApp/TestApp.cpp
toddbellMSFT: We will not be pursuing a targeted fix for GDK. I will update the comment.

source/test/TestApp/TestApp.cpp
toddbellMSFT: Yeah, Intellisense was saying it was still being compiled, after I made this fix things ran just fine.

PR #127, opened 2021-11-11
Initial Connection String

toddbellMSFT:
Added general tests for swapping out an estimation of the connection string.


source/code/source/playfab/PlayFabApiSettings.cpp
```diff
@@ -20,12 +20,24 @@ namespace PlayFab
...
+            fullUrl += "https://";
+            fullUrl += titleId;
+            fullUrl += baseServiceHost;
+            fullUrl += urlPath;
+        }
+        else
+        {
+            if(PlayFabSettings::connectionString == "")
+            {
+                PlayFabSettings::connectionString = connectionString;
```
amccalib, line 36: I don't think we want to do this. My understanding is that PlayFabSettings is the static, program-wide configuration while PlayFabApiSettings is instanced and could potentially apply to a specific set of calls only. So, instead of setting the static variable when we get an instanced connection string, we should just treat it as an instanced override.

amccalib (changes requested):

source/code/source/playfab/PlayFabApiSettings.cpp
toddbellMSFT: Sure thing. I was trying to make it work with both. But since the static settings is considered legacy behavior, it would make a smaller change to apply across the remaining sdks that behave similarly.

toddbellMSFT: Updated to check instance first, then defer to static if present. Then build as we usually have.

amccalib (approved):

PR #131, opened 2022-04-08
Generator changes to create azure flavor

AlonsoMondal:


source/build/Windows/TestClientApp.vcxproj.ejs
```diff
@@ -62,7 +62,7 @@
       <WarningLevel>Level4</WarningLevel>
       <Optimization>Disabled</Optimization>
       <SDLCheck>true</SDLCheck>
-      <PreprocessorDefinitions>_DEBUG;_CONSOLE;WIN32;<%- clientDefines %>%(PreprocessorDefinitions)</PreprocessorDefinitions>
+      <PreprocessorDefinitions>_DEBUG;_CONSOLE;WIN32;<% if(azureSdk){ %> <%- azureDefines %> <% } %><%- clientDefines %>%(PreprocessorDefinitions)</PreprocessorDefinitions>
```
pgilmorepf, line 65: NIT: Use azureDefines in the way that clientDefines is used, and remove the need for azureSdk.

source/build/Windows/TestClientApp.vcxproj.ejs
```diff
@@ -86,7 +86,7 @@
       <PrecompiledHeader>Use</PrecompiledHeader>
       <WarningLevel>Level4</WarningLevel>
       <SDLCheck>true</SDLCheck>
-      <PreprocessorDefinitions>_CONSOLE;WIN32;<%- clientDefines %>%(PreprocessorDefinitions)</PreprocessorDefinitions>
+      <PreprocessorDefinitions>_CONSOLE;WIN32;<% if(azureSdk){ %> <%- azureDefines %> <% } %><%- clientDefines %>%(PreprocessorDefinitions)</PreprocessorDefinitions>
```
pgilmorepf, line 89: NIT: Use azureDefines in the way that clientDefines is used, and remove the need for azureSdk.

source/code/source/playfab/PlayFabEventPipeline.cpp
```diff
@@ -354,24 +354,12 @@ namespace PlayFab
...
-            eventsApi->WriteEvents(
-                batchReq,
-                std::bind(&PlayFabEventPipeline::WriteEventsApiCallback, this, std::placeholders::_1, std::placeholders::_2),
-                std::bind(&PlayFabEventPipeline::WriteEventsApiErrorCallback, this, std::placeholders::_1, std::placeholders::_2),
-                customData);
-        }
-        else
-        {
-            // call Events API to send the batch, bypassing Playstream
-            eventsApi->WriteTelemetryEvents(
```
pgilmorepf, line 369: Once upon a time, this separation was deliberate, and quite important. TelemetryEvents don't do all the same PlayStream post-processing, and are therefore SIGNIFICANTLY faster in many ways.

pgilmorepf (approved):
Approved with nitpicks.

PLZ confirm that you're really removing the TelemetryEvents for this PR. That was a very important feature way back when.

source/code/include/playfab/PlayFabApiSettings.h
```diff
@@ -23,6 +23,7 @@ namespace PlayFab
         std::string baseServiceHost; // The base for a PlayFab service host
         std::string titleId; // You must set this value for PlayFabSdk to work properly (found in the Game Manager for your title, at the PlayFab Website)
         std::string connectionString; // set to override minimal string building from the client. Produced directly from the game manager.
+        std::string playerAccountPoolId; // TODO: appropriate description
```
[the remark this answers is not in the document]
amccalib: Also, is there anything that keeps this from showing up in non-azure SDKs? In the other SDKs, I had to turn this into an ejs file and conditionally check the azureSdk flag.

source/code/include/playfab/PlayFabAuthenticationContext.h
```diff
@@ -18,6 +18,7 @@ namespace PlayFab
         std::string clientSessionTicket; // Client session ticket that is used as an authentication token in many PlayFab API methods.
 
     public:
+        std::string playerAccountPoolId;
```
amccalib, line 21: Similar to above, what keeps this from showing up in non-Azure SDKs?

source/code/source/playfab/PlayFabApiSettings.cpp
```diff
@@ -14,6 +14,7 @@ namespace PlayFab
         if (PlayFabSettings::staticSettings != nullptr)
         {
             titleId = PlayFabSettings::staticSettings->titleId;
+            playerAccountPoolId = PlayFabSettings::staticPlayer->playerAccountPoolId;
```
amccalib, line 17: Should only be in Azure SDKs.

source/code/source/playfab/PlayFabAuthenticationContext.cpp
```diff
@@ -15,6 +15,7 @@ namespace PlayFab
         playFabId.clear();
         clientSessionTicket.clear();
 #endif
+        playerAccountPoolId.clear();
```
amccalib, line 18: Only in Azure SDKs.

amccalib (changes requested):

source/code/source/playfab/PlayFabEventPipeline.cpp
amccalib: I'm not sure why this is being removed. Please explain.

source/code/source/playfab/QoS/PlayFabQoSApi.cpp
```diff
@@ -182,7 +182,7 @@ namespace PlayFab
             eventContents.Payload = value;
             request.Events.push_back(eventContents);
 
-            eventsApi->WriteTelemetryEvents(request, WriteEventsSuccessCallBack, WriteEventsFailureCallBack);
+            eventsApi->WriteEvents(request, WriteEventsSuccessCallBack, WriteEventsFailureCallBack);
```
amccalib, line 185: Why this change? Is the WriteTelemetryEventsAPI missing from Azure? We should discuss more, if so, as that may be in error.

source/code/source/playfab/PlayFabEventPipeline.cpp
AlonsoMondal: I incorrectly assumed that WriteEvents & WriteTelemetryEvents were consolidated in the cerulean versions as the latter is not included in the cerulean APIs. I have filed a bug to track this: https://dev.azure.com/microsoft/Xbox/_sprints/taskboard/GDX.SDKs.Core/Xbox/22-4W20%20(Ends%205-13)/2W18%20(Ends%204-29)?workitem=39141518

once this is solved, I'll get back to adding this once again

templates/PlayFabAuthenticationContext.h.ejs
```diff
@@ -1,37 +1,40 @@
...
+        std::string entityToken; // User's entity token. Entity tokens are required by all Entity API methods.
+
+        PlayFabAuthenticationContext();
+        PlayFabAuthenticationContext(const PlayFabAuthenticationContext& other) = delete;
+        PlayFabAuthenticationContext(PlayFabAuthenticationContext&& other) = delete;
+        PlayFabAuthenticationContext& operator=(const PlayFabAuthenticationContext& other) = delete;
+        PlayFabAuthenticationContext& operator=(PlayFabAuthenticationContext&& other) = delete;
+        ~PlayFabAuthenticationContext() = default;
+
+        void HandlePlayFabLogin(const std::string& _playFabId, const std::string& _clientSessionTicket, const std::string& _playerAccountPoolId, const std::string& _entityId, const std::string& _entityType, const std::string& _entityToken);
```
amccalib, line 35: You'll need two versions of HandlePlayFabLogin depending on whether AzureSdk is true. One with _playerAccountPoolId and one without.

amccalib (changes requested):

make.js
```diff
@@ -1,324 +1,390 @@
...
+            tabbing + "    authKey = \"X-SecretKey\"; authValue = settings->developerSecretKey;\n" +
+            tabbing + "}\n" +
+            "#endif\n";
+
+    return "";
+}
+
+function getResultActions(tabbing, apiCall, isInstanceApi) {
+    if (apiCall.url === "/Authentication/GetEntityToken")
+        return tabbing + "context->HandlePlayFabLogin(\"\", \"\",  \"\", outResult.Entity->Id, outResult.Entity->Type, outResult.EntityToken);\n";
```
amccalib, line 332: I think these calls to HandlePlayFabLogin need to be updated as well. There is one less parameter in the non-Azure cases.

templates/PlayFabAuthenticationContext.h.ejs
```diff
@@ -1,37 +1,40 @@
...
+        std::string entityToken; // User's entity token. Entity tokens are required by all Entity API methods.
+
+        PlayFabAuthenticationContext();
+        PlayFabAuthenticationContext(const PlayFabAuthenticationContext& other) = delete;
+        PlayFabAuthenticationContext(PlayFabAuthenticationContext&& other) = delete;
+        PlayFabAuthenticationContext& operator=(const PlayFabAuthenticationContext& other) = delete;
+        PlayFabAuthenticationContext& operator=(PlayFabAuthenticationContext&& other) = delete;
+        ~PlayFabAuthenticationContext() = default;
+        
+        void HandlePlayFabLogin(const std::string& _playFabId, const std::string& _clientSessionTicket, <% if(azureSdk) { %>const std::string& _playerAccountPoolId,<% } %> const std::string& _entityId, const std::string& _entityType, const std::string& _entityToken);
```
amccalib, line 35: Thinking on this a bit more, do we even need to be taking a parameter here for the player account pool? That should be set before making a call to login, not after. The purpose of this method is to "cache" the auth context from the login call, which means the player account pool id should already have been set.

amccalib (changes requested):

amccalib (approved):

make.js
```diff
@@ -1,324 +1,390 @@
...
+        sdkVersion: sdkGlobals.sdkVersion,
+        sdkDate: sdkGlobals.sdkVersion.split(".")[2],
+        sdkYear: sdkGlobals.sdkVersion.split(".")[2].substr(0, 2),
+        vsVer: "v141", // As C++ versions change, we may need to update this
+        vsYear: "2017", // As VS versions change, we may need to update this
+        getVerticalNameDefault: getVerticalNameDefault,
+        winSdkVersion: "10.0.17763.0" // Which version of the Windows SDK (A VS installation option) to use
+    };
+
+    if(sdkGlobals.buildFlags.includes("azure")){
```
amccalib, line 33: This check is duplicated down below in isAzureSDK. Maybe just turn this check into a call to that method. So, something like:

if (isAzureSdk()){
...
}

amccalib (approved):

PR #133, opened 2022-05-18
Tb/merge gdk template

toddbellMSFT:
Adding in initial GDK template changes that are compatible with the windows builds.


source/build/Gdk/MainWin32.cpp
```diff
@@ -0,0 +1,11 @@
+// Copyright (C) Microsoft Corporation. All rights reserved.
+
+#include "TestAppPch.h"
+#include "TestApp.h"
```
jasonsandlin, line 4: source\build\gdk seems odd place for test code.  i haven't looked at full project layout but consider moving test app code to a test folder. eg something like source\tests\gdk or whatever makes sense.

jasonsandlin (approved):
:shipit:

source/code/include/playfab/PlayFabPlatformMacros.h
```diff
@@ -9,6 +9,10 @@
 
 #pragma once
 
+// TODO: is there a replacement that we can test instead?
+#ifdef _GAMING_DESKTOP
```
amccalib, line 13: Reflecting on this, I suspect this is only defined on Windows GDK builds, not Console. @jasonsandlin do you know if there's a good token to determine the presence of the GDK on Console?

source/code/source/playfab/PlayFabCurlHttpPlugin.cpp
```diff
@@ -133,11 +139,15 @@ namespace PlayFab
             return nitems * size; // The return expected by curl for this callback
         }
 
+size_t headerKeyLength = size;
+#if defined(PLAYFAB_PLATFORM_GDK)
+        headerKeyLength = nitems;
+#endif
         // If this header-line is long enough, and the header starts with the key we expect
-        if ((nitems > requestIdheaderKeyLen) && (strncasecmp(buffer, requestIdHeaderKey, requestIdheaderKeyLen) == 0))
+        if ((headerKeyLength > requestIdheaderKeyLen) && (strncasecmp(buffer, requestIdHeaderKey, requestIdheaderKeyLen) == 0))
```
amccalib, line 147: I'm a little worried this is a change in behavior for non-GDK. Previously, for non-GDK, we would compare nitems to requestIdheaderKeyLen. Now, though, for non-GDK, we effectively compare size to requestIdheaderKeyLen. Does the #if above need to be an #ifndef

source/code/source/playfab/PlayFabEvent.cpp
```diff
@@ -65,9 +65,11 @@ namespace PlayFab
         this->eventContents.Payload[name] = (Int64)value;
 #elif defined(JSON_HAS_INT64)
         this->eventContents.Payload[name] = (Json::Int64)value;
-#else // PLAYFAB_PLATFORM_ANDROID
+#elif defined(PLAYFAB_PLATFORM_GDK)
+        this->eventContents.Payload[name] = (Json::Int)value;
+#else
```
amccalib, line 70: Should the // PLAYFAB_PLATFORM_ANDROID be added back here?

source/code/source/playfab/PlayFabEvent.cpp
```diff
@@ -93,9 +95,11 @@ namespace PlayFab
         this->eventContents.Payload[name] = (Uint64)value;
 #elif defined(JSON_HAS_INT64)
         this->eventContents.Payload[name] = (Json::UInt64)value;
-#else // PLAYFAB_PLATFORM_ANDROID
+#elif defined(PLAYFAB_PLATFORM_GDK)
+        this->eventContents.Payload[name] = (Json::UInt)value;
+#else
```
amccalib, line 100: Same as above. Add // PLAYFAB_PLATFORM_ANDROID back?

amccalib (changes requested):

source/code/include/playfab/PlayFabPlatformMacros.h
jasonsandlin: libHttpClient does this:
    #if defined(_GAMING_DESKTOP) || defined(_GAMING_XBOX) || defined(_GAMING_XBOX_XBOXONE) || defined(_GAMING_XBOX_SCARLETT)
if any of those 3 are defined, then we infer its a GDK build.  

for this project, if you want PC GDK builds only use 
#if defined(_GAMING_DESKTOP)

If you want console GDK builds only  then use
#if defined(_GAMING_XBOX) || defined(_GAMING_XBOX_XBOXONE) || defined(_GAMING_XBOX_SCARLETT)

if you want PC & console GDK builds, then use all 4 or'd

source/code/source/playfab/PlayFabCurlHttpPlugin.cpp
toddbellMSFT: I'm seeing the error that caused this swap is no longer triggering, I'll remove the change.

source/code/include/playfab/PlayFabPlatformMacros.h
amccalib: We want to support both with this, so we'll need all four.

PR #144, opened 2022-07-21
nuspec file update

toddbellMSFT:
C++ nuget packaging is running into legacy compatability issues


source/com.playfab.xplatcppsdk.v141.nuspec.ejs
```diff
@@ -22,6 +22,12 @@
         <file src=".\build\Windows\x64\$configuration$\XPlatCppWindows\XPlatCppWindows.lib" target="build\native\lib\x64\$configuration$\XPlatCppWindows.lib" />
         <file src=".\build\Windows\x64\Debug\XPlatCppWindows\XPlatCppWindows.pdb" target="build\native\lib\x64\Debug\XPlatCppWindows.pdb" />
         <file src=".\build\Windows\x64\Debug\XPlatCppWindows\XPlatCppWindows.lib" target="build\native\lib\x64\Debug\XPlatCppWindows.lib" />
+        <file src=".\external\jsoncpp-gdk-build\Gaming.Desktop.x64\$configuration$\lib_json.lib" target="build\native\lib\Gaming.Desktop.x64\$configuration$\lib_json.lib" />
```
amccalib, line 25: I know the existing file items use the same %configuration% token, but I don't understand why. The lines immediately below directly uses "Debug". Would %configuration% ever be anything other than Release? What if it was Debug too?

toddbellMSFT: https://stackoverflow.com/questions/40628116/how-to-specify-configuration-specific-folder-in-nuspec#:~:text=In%20the.nuspec%20file%20reference%2C%20under%20Replace%20Tokens%2C%20it,always%20use%20-properties%20Configuration%3DRelease%20on%20the%20command%20line.%22 we are calling Configuration=Release in the ADO build pipeline when we call nuget pack. I tried hard coding it in the past, but nuspec was complaining about a lack of a $configuration$ token. I can try it again so we don't need to rely on another possible changing surface area when we know exactly what we want to build/package/deploy

amccalib: No, I'm fine leaving it. Just a weird artifact. Chances of it biting us in the future are very slim.

amccalib (approved):

PR #147, opened 2022-08-12
Update make.js

toddbellMSFT:


toddbellMSFT:
We took a different approach for ejs. This make.js update should not be needed.

PR #151, opened 2022-10-10
Auth GameServer EntityToken caching & deleting

AlonsoMondal:
This PR goes over the changes to cache the result of AuthenticateGameServerWithCustomId in a seperate "game server entity token".

Request actions for this specific function should be reviewed, not sure if they should have the same options as GetEntityToken but changes are as if they are. Might want to change the default EntityToken for the new "GameServerEntityToken" when setting request actions.

Deleting sets the HandleLogin to empty strings.

### DIFF
--- a/make.js
+++ b/make.js
@@ -152,7 +152,7 @@ function hasAuthParams(apiCall) {
 }
 
 function getAuthParams(apiCall, isInstanceApi) {
-    if (apiCall.url === "/Authentication/GetEntityToken"  || apiCall.url === "/GameServerIdentity/AuthenticateGameServerWithCustomId")
+    if (apiCall.url === "/Authentication/GetEntityToken")
         return "authKey, authValue";
     switch (apiCall.auth) {
         case "EntityToken": return "\"X-EntityToken\", context->entityToken";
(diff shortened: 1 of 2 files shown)


toddbellMSFT (approved):

source/code/source/playfab/PlayFabAuthenticationContext.cpp.ejs
```diff
@@ -45,6 +45,21 @@
             SetIfNotNull(_entityType, entityType);
             SetIfNotNull(_entityToken, entityToken);
         }
+
+        void PlayFabAuthenticationContext::HandlePlayFabGameServerLogin(
```
amccalib, line 49: Is this needed? It looks identical to the method above (HandlePlayFabLogin).

make.js
```diff
@@ -339,6 +339,10 @@ function getRequestActions(tabbing, apiCall, isInstanceApi) {
 function getResultActions(tabbing, apiCall, isInstanceApi) {
     if (apiCall.url === "/Authentication/GetEntityToken")
         return tabbing + "context->HandlePlayFabLogin(\"\", \"\", outResult.Entity->Id, outResult.Entity->Type, outResult.EntityToken);\n";
+    if (apiCall.url === "/GameServerIdentity/AuthenticateGameServerWithCustomId")
+        return tabbing + "context->HandlePlayFabGameServerLogin(\"\", \"\", outResult.EntityToken->Entity->Id, outResult.EntityToken->Entity->Type, outResult.EntityToken->EntityToken);\n";
+    if (apiCall.url === "/GameServerIdentity/Delete")
```
amccalib, line 344: I don't think we should add any special handling for Delete. We risk clearing a valid entity token in some edge cases (like if the cached entity token had been changed back to a title entity token, or another game_server)

make.js
```diff
@@ -152,7 +152,7 @@ function hasAuthParams(apiCall) {
 }
 
 function getAuthParams(apiCall, isInstanceApi) {
-    if (apiCall.url === "/Authentication/GetEntityToken")
```
amccalib, line 155: Like the other PR, the new API doesn't need any special auth handling.

amccalib (changes requested):

amccalib (approved):

PR #152, opened 2022-11-29
Update XPlatSocket.cpp

toddbellMSFT:
Split set sockopt calls to two separate calls. There is a possibility that windows is not bubbling out an error when we set using a bitwise or.


toddbellMSFT:
This should cover the changes in https://github.com/PlayFab/XPlatCppSdk/pull/49/

amccalib (approved):

PR #153, opened 2022-12-15
Batch up remaining events before send batch on shutdown

boczar:
Any events left in the buffer would get lost instead of getting sent.

### DIFF
--- a/source/code/source/playfab/PlayFabEventPipeline.cpp
+++ b/source/code/source/playfab/PlayFabEventPipeline.cpp
@@ -225,6 +225,15 @@ namespace PlayFab
                 std::this_thread::sleep_for(std::chrono::milliseconds(this->settings->readBufferWaitTime));
             }
         }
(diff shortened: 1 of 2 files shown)


amccalib (approved):

source/code/source/playfab/PlayFabEventPipeline.cpp
```diff
@@ -225,6 +225,15 @@ namespace PlayFab
                 std::this_thread::sleep_for(std::chrono::milliseconds(this->settings->readBufferWaitTime));
             }
         }
+        // batch up any buffered events
+        while( PopBuffer() )
```
amccalib, line 229: Just ran this through a generate/build real quick. PopBuffer needs to be added to PlayFabEventPipeline.h as well.

amccalib (changes requested):

amccalib (approved):

